PR #2, opened 2018-05-23
Distributed Supervisor backed by δ-CRDT

derekkraan:


lib/horde/supervisor.ex
```diff
@@ -0,0 +1,423 @@
...
+        {_id, {^dead_node, _child_spec}} -> true
+        _ -> false
+      end)
+      |> Enum.filter(fn {_id, {_node, child}} ->
+        case choose_node(child.id, state) do
+          {^this_node_id, _node_info} -> true
+          _ -> false
+        end
+      end)
+      |> Enum.map(fn {_id, {_node, child}} -> add_child(child, state) end)
```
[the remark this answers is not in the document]
Tuxified: BTW both on line 356 as in add_child/2 we call choose_node/2 which uses a hash function. I don't know how costly that function is, but perhaps we should reduce the calls to choose_node/hash function?

lib/horde/supervisor.ex
```diff
@@ -0,0 +1,423 @@
...
+  end
+
+  defp claim_unclaimed_processes(state) do
+    this_node_id = state.node_id
+
+    state.processes
+    |> Enum.each(fn
+      {id, {nil, child_spec}} ->
+        case choose_node(id, state) do
+          {^this_node_id, _node_info} -> add_child(child_spec, state)
```
Tuxified: both this and previous line call into choose_node/2, will this make any impact on performance?

lib/horde/supervisor.ex
```diff
@@ -0,0 +1,423 @@
...
+
+      :ok = Supervisor.stop(state.supervisor_pid)
+
+      send(horde, :force_shutdown)
+    end)
+  end
+
+  defp handle_updated_members_pids(state, new_state) do
+    new_pids =
+      Enum.map(new_state.members, fn {_key, {_, _, _, m, _}} -> m end) |> Enum.into(MapSet.new())
```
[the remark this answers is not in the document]
derekkraan: (maybe) even better: `MapSet.new/2`

lib/horde/supervisor.ex
derekkraan: The biggest performance problems right now are in the CRDT, I'll be focussing on that first.

lib/horde/supervisor.ex
```diff
@@ -308,11 +308,8 @@ defmodule Horde.Supervisor do
   end
 
   defp handle_updated_members_pids(state, new_state) do
-    new_pids =
-      Enum.map(new_state.members, fn {_key, {_, _, _, m, _}} -> m end) |> Enum.into(MapSet.new())
-
-    old_pids =
-      Enum.map(state.members, fn {_node_id, {_, _, _, m, _}} -> m end) |> Enum.into(MapSet.new())
+    new_pids = MapSet.new(new_state.members, fn {_key, {_, _, _, m, _}} -> m end)
+    old_pids = MapSet.new(state.members, fn {_key, {_, _, _, m, _}} -> m end)
```
Tuxified: 👍

PR #3, opened 2018-05-29
A bunch of changes to get closer to production-ready

derekkraan:


lib/horde/supervisor.ex
```diff
@@ -274,6 +305,7 @@ defmodule Horde.Supervisor do
     end
   end
 
+  @doc false
```
Tuxified: private functions don't need @docs

test/horde_supervisor_test.exs
```diff
@@ -38,7 +38,7 @@ defmodule HordeSupervisorTest do
 
   describe ".start_child/2" do
     test "starts a process", context do
-      Horde.Supervisor.start_child(context.horde_1, context.task_def)
+      assert {:ok, _pid} = Horde.Supervisor.start_child(context.horde_1, context.task_def)
```
Tuxified: if we know the `pid`, would be a nice extra check it's also the one we receive at [line 43](https://github.com/derekkraan/horde/pull/3/files#diff-a8d0ac8f48d80df61cea093956142e33R43) ?

lib/horde/supervisor.ex
```diff
@@ -25,8 +25,16 @@ defmodule Horde.Supervisor do
               shutting_down: false
   end
 
-  def child_spec(id) do
-    %{id: id, start: {__MODULE__, :start_link, [id]}}
+  def child_spec(options \\ []) do
+    options =
+      Keyword.put_new(options, :id, __MODULE__)
+      |> Keyword.put_new(:children, [])
```
Tuxified: Not important, but I find this style a bit more readable
```
options =
  options
  |> Keyword.put_new(:id, __MODULE__)
  |> Keyword.put_new(:children, [])
```

Tuxified: perhaps use Keyword.merge/2:
`options = Keyword.merge([id: __MODULE__, children: []], options)`

lib/horde/registry.ex
```diff
@@ -18,11 +18,18 @@ defmodule Horde.Registry do
...
-  def child_spec(node_id) do
-    %{id: node_id, start: {GenServer, :start_link, [__MODULE__, node_id]}}
+  def child_spec(options \\ []) do
+    options = Keyword.put_new(options, :id, __MODULE__)
+
+    %{
+      id: options[:id],
+      start:
+        {GenServer, :start_link,
+         [__MODULE__, Keyword.delete(options, :id), Keyword.take(options, [:name])]}
```
Tuxified: What if someone forgets to supply a `name:` option? Should we guard against it?

derekkraan: Good question. What does `Registry` do?

derekkraan: I looked it up: `Registry` uses `name` as the `id` of the process in the `child_spec`. I'll duplicate this behaviour in `Horde.Registry`.

PR #7, opened 2018-06-14
Use ETS to remove bottleneck on Horde.Registry read operations

derekkraan:


lib/horde/registry.ex
```diff
@@ -172,7 +184,15 @@ defmodule Horde.Registry do
...
     processes = GenServer.call(state.processes_pid, {:read, @crdt})
 
-    {:noreply, %{state | processes: processes}}
+    :ets.insert(state.ets_table, Map.to_list(processes))
+
+    all_keys = :ets.match(state.ets_table, {:"$1", :_}) |> MapSet.new(fn [x] -> x end)
+    new_keys = Map.keys(processes) |> MapSet.new()
+    to_delete_keys = MapSet.difference(all_keys, new_keys)
+
+    to_delete_keys |> Enum.each(fn key -> :ets.delete(state.ets_table, key) end)
```
Tuxified: I wonder if there isn't a way to delete multiple items in 1 go?

derekkraan: Yeah, all items at once ;)

PR #12, opened 2018-07-03
Remove `Horde.Cluster.leave_hordes` & some other clean up

derekkraan:


lib/horde/supervisor.ex
```diff
@@ -2,23 +2,34 @@ defmodule Horde.Supervisor do
...
+
+  ## Divergence from standard DynamicSupervisor behaviour
+
+  While Horde wraps DynamicSupervisor, it does keep track of processes by the `id` in the child specification. This is a divergence from the behaviour of DynamicSupervisor, which ignores ids altogether. Using DynamicSupervisor is useful for its shutdown behaviour (it shuts down all child processes simultaneously, unlike `Supervisor`).
+
+  ## Graceful shutdown
+
+  Horde migrates processes from one node to another when you stop a node. The state of child processes is not preserved, the process is simply restarted on another node.
```
Tuxified: A node can be stopped by various means, simply saying processes will be transferred/migrated when a node stops gives the impression there's some magic going on. Perhaps rephrase this? Ex.: `When a node stops, horde restarts the processes of stopped node on another node. The state.. `

lib/horde/supervisor.ex
```diff
@@ -155,49 +165,38 @@ defmodule Horde.Supervisor do
...
-        {:reply, error, state}
+      {other_node, _child_spec} ->
+        proxy_to_node(other_node, msg, from, state)
     end
   end
 
   @doc false
   def handle_call({:start_child, _child_spec}, _from, %{shutting_down: true} = state),
-    do: {:reply, {:error, "shutting down"}, state}
+    do: {:reply, {:error, {:shutting_down, "this node is shutting down. please retry."}}, state}
```
Tuxified: `please retry elsewhere`? if a node is shutting down, folks shouldn't "just retry" :P

lib/horde/cluster.ex
```diff
@@ -11,27 +11,16 @@ defmodule Horde.Cluster do
...
+  Join two hordes into one big horde. Calling this once will inform every node in each horde of every node in the other horde. Leave the hordes by calling `Horde.Supervisor.stop/1` or `Horde.Registry.stop/1`
   """
-  @spec leave_hordes(horde :: pid()) :: :ok
-  def leave_hordes(horde) do
-    GenServer.cast(horde, :leave_hordes)
+  @spec join_hordes(horde :: pid(), other_horde :: pid()) :: boolean()
+  def join_hordes(horde, other_horde, timeout \\ 5000) do
+    GenServer.call(horde, {:join_hordes, other_horde}, timeout)
+  catch
+    :exit, {:timeout, _details} -> false
```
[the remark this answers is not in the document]
derekkraan: I will add a log message. To keep the interface consistent with `Node.connect`, the return value will stay the same.

lib/horde/registry.ex
```diff
@@ -54,16 +54,36 @@ defmodule Horde.Registry do
...
+    GenServer.stop(registry, reason, timeout)
+  end
+
+  def terminate(reason, state) do
+    GenServer.cast(
+      state.members_pid,
+      {:operation, {@crdt, :remove, [state.node_id]}}
+    )
+
+    Kernel.send(state.members_pid, :ship_interval_or_state_to_all)
```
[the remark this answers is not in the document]
derekkraan: Actually good point, I can remove this line because this is now automatic shutdown behaviour in `DeltaCRDT`

lib/horde/supervisor.ex
derekkraan: Removed retry part, we can't know why a node is shutting down.

Issue #11, opened 2018-06-29
Issues on joining/leaving hordes 


mszmurlo:
Hi,

Unless I misunderstood something, I believe there are some issues in
joining/leaving hordes. Below are the traces and comments on
experiment I have done so far. As all is about the same topic, I open just
one issue

For what follows I have a `XYZ.Utils` module with two helper functions that will be used later:

    def start_worker(name) do
      cs = %{id: name, start: {XYZ.Worker, :start_link, [[client_id: name]]}}
      Horde.Supervisor.start_child(XYZ.DSup, cs)
    end

    def lookup(name) do
     Horde.Registry.lookup(XYZ.Worker.via_tuple(name))
    end

and a `XYZ.Worker` module which is basically a `GenServer` with few
API that are not relevant here.

All examples start with a fresh situation with two nodes started with
`ERL_AFLAGS="-name nd1@127.0.0.1 -setcookie abcd" iex -S mix` and
`ERL_AFLAGS="-name nd2@127.0.0.1 -setcookie abcd" iex -S mix` in two
terminals. Then `nd1` connect to `nd2` with
`Node.connect(:"nd2@127.0.0.1")`.

## I1 : joining a horde on a node that does not exist

Joining the hordes from `nd1` to `nd2` works fine

    iex(nd1@127.0.0.1)1> Horde.Cluster.join_hordes(XYZ.DSup, {XYZ.DSup, :"nd2@127.0.0.1"})
    :ok
    iex(nd1@127.0.0.1)2> Horde.Cluster.join_hordes(XYZ.DReg, {XYZ.DReg, :"nd2@127.0.0.1"})
    :ok

However, the system returns `:ok` if you specify a node name that does not exist:

    iex(nd1@127.0.0.1)9> Horde.Cluster.join_hordes(XYZ.DReg, {XYZ.DReg, :"nd4@127.0.0.1"})
    :ok

Same holds for supervisors. Returning `false` would provide the same
semantics as `Node.connect()`.

## I2: Upon `leave_hordes()` registry on removed node is not cleaned properly

Start with a fresh situation, join the hordes and start a process
that will run on `nd1`. Check on both nodes that the process is available:

    iex(nd1@127.0.0.1)11> XYZ.Utils.lookup("a1")
    #PID<0.215.0>

and on `nd2`

    iex(nd2@127.0.0.1)5> XYZ.Utils.lookup("a1")
    #PID<18754.215.0>

[...]

PR #15, opened 2018-07-24
Fix connection issue

derekkraan:
Fix reconnection issue.

What was happening:
If you connected to a horde, disconnected, and then connected again, things were going wrong.
Internally, the delta_crdt maintains a list of neighbours, and a map of neighbours to ack'ed deltas. The reference used for a neighbour was in this case of the form `{HelloWorld.HelloSupervisor, :"hello1@127.0.0.1"}` (in the case of the example app). The problem is: if this process were to die and be restarted, the reference would not change, so the neighbours would not send in the correct deltas.

The solution is to only communicate using pids. So now the neighbours list is a simple list of pids. This ensures that when these processes (or the entire node) are restarted, that the reference in the neighbours list is new.

### DIFF
--- a/lib/horde/registry.ex
+++ b/lib/horde/registry.ex
@@ -282,12 +282,7 @@ defmodule Horde.Registry do
     {:reply, {:ok, state.members}, state}
   end
 
-  defp generate_node_id(bits \\ 128) do
-    <<num::bits>> =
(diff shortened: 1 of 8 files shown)


derekkraan:
fixes #11

PR #27, opened 2018-09-29
WIP Registry arbitrary terms

derekkraan:


derekkraan:
I think the key to this is going to be emulating the ETS table structure of stdlib's `Registry` and also mirroring that in the same number of delta CRDTs.

I thought this would be simple so I opened a WIP PR right away but having done some reading of the source code I suspect this will take a bit longer. Putting on hold for now.

Issue #28, opened 2018-10-02
Unable to start child if child_spec.id doesn't implement String.Chars


amatalai:
Example:

```elixir
iex(1)> child_spec = %{id: {A, "a"}, start: {Agent, :start_link, [fn -> 0 end]}}

iex(2)> DynamicSupervisor.start_child(MyApp.DynamicSupervisor, child_spec)
{:ok, #PID<0.836.0>}

iex(3)> Horde.Supervisor.start_child(MyApp.DistributedSupervisor, child_spec)
[error] GenServer MyApp.DistributedSupervisor terminating
** (Protocol.UndefinedError) protocol String.Chars not implemented for {A, "a"}. This protocol is implemented for...
```

This is problematic when worker definition is provided by external lib

PR #29, opened 2018-10-02
Allow Supervisor to start process with arbitrary term in childspec id

amatalai:
Attempt to solve #28


derekkraan:
👏👏 thank you!

PR #30, opened 2018-10-02
fixed dialyzer warnings caused by Horde.Cluster specs

amatalai:
When using module name instead of pids, example:
```elixir
Enum.each(Node.list(), fn node ->
  Horde.Cluster.join_hordes(
    MyApp.DistributedRegistry,
    {MyApp.DistributedRegistry, node}
  )
end)
```
I encountered following dialyzer warnings (using dialyxir 1.0.0-rc.3)
```
lib/my_app/horde_connector.ex:29:call
The call:
Horde.Cluster.join_hordes(MyApp.DistributedRegistry, {MyApp.DistributedRegistry, _})

breaks the contract
(horde :: pid(), other_horde :: pid()) :: boolean()
```

This PR fixes this.

### DIFF
--- a/lib/horde/cluster.ex
+++ b/lib/horde/cluster.ex
@@ -18,7 +18,7 @@ defmodule Horde.Cluster do
   @doc """
   Join two hordes into one big horde. Calling this once will inform every node in each horde of every node in the other horde. Leave the hordes by calling `Horde.Supervisor.stop/1` or `Horde.Registry.stop/1`
   """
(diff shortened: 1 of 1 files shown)


derekkraan:
Could we lean on the `Genserver.server()` type for this one?

amatalai:
Yes. `GenServer.server()` is valid typespec (althrough `Horde.Cluster.join_hordes({:global, A}, {:global, B})` would be very weird thing to do). Pushed fix.

PR #34, opened 2018-11-08
Clean pids from Registry when process dies

### LINKED ISSUE #16: Registries not cleaned properly after node crash or shutdown

derekkraan:
- Clean pids from registry when they die
- Include mechanism to ensure that rejoining registries after a net split doesn't wreck the registry.

Fixes #16


lib/horde/registry_impl.ex
```diff
@@ -121,26 +142,35 @@ defmodule Horde.RegistryImpl do
...
+        end)
+
     member_pids =
-      MapSet.new(members, fn {_key, {members_pid, _registry_pid, _keys_pid}} ->
+      MapSet.new(members, fn {_key, %{members_pid: members_pid}} ->
         members_pid
       end)
       |> MapSet.delete(nil)
 
+    monitor_members(members, state)
```
Tuxified: because of the name I thought this would start monitoring all members every time the members updated occured. perhaps `monitor_new_members` is a more clear name?

derekkraan: It is clearer. Will change.

PR #41, opened 2018-11-23
Update xxhash version 0.2.0 => 0.2.1 for the example "Hello World" application. 

Ivor:
The example app was referencing a previous version of xxhash which contained a bug that was fixed in the latest version. I discovered this while trying to run the hello world application. 

Updating to the latest version fixes the issue. 
Thanks for a great library. I am excited to play with it.


derekkraan:
Thanks!

PR #44, opened 2018-11-26
Update example app to share data across the Horde.

Ivor:
In the example we initialize the state of the `HelloWorld.SayHello` GenServer with the "count" value stored in `Horde.Registry` meta data. When `:say_hello` is received, we now also increment the "count" value. 

If the node running the single `HelloWorld.SayHello` instance is killed, and the supervisor starts a new instance, the state is read from the shared meta data and the counter continues where the previous instance of `SayHello` left off.


Ivor:
Thanks for the help in knowing where to look. Happy to accommodate feedback to make this better.

derekkraan:
Thanks for the contribution! Looks good to me!

PR #48, opened 2018-12-05
Pass supervisor options through to SupervisorImpl

Ivor:
I noticed that the default distribution_strategy in `SupervisorImpl` is Horde.UniformDistribution and that setting the `distritbution_strategy` in the options to `Horde.Supervisor` was not being passed through to allow one to override the `distribution_strategy` with something else. 

I think this change will allow that. It works for me locally.


derekkraan:
Thanks! Must have overlooked that in a refactor.

Issue #49, opened 2018-12-05
Quorum based distribution fails when calling `choose_node` on `start_child`


Ivor:
The `UniformQuorumDistribution` strategy, and any similar ones where `has_quorum?/1` can return `false` and `choose_node` therefore `nil` breaks the `:start_child` call in `SupervisorImpl`.

```
defmodule Horde.UniformQuorumDistribution do
  def choose_node(identifier, members) do
    if has_quorum?(members) do
      Horde.UniformDistribution.choose_node(identifier, members)
    else
      nil
    end
  end
```
and
```
defmodule Horde.SupervisorImpl do
  def handle_call({:start_child, child_spec} = msg, from, %{node_id: this_node_id} = state) do
    case state.distribution_strategy.choose_node(child_spec.id, state.members) do
      {^this_node_id, _} ->
        {reply, new_state} = add_child(child_spec, state)
        {:reply, reply, new_state}

      {other_node_id, _} ->
        proxy_to_node(other_node_id, msg, from, state)
    end
  end
```

If a cluster is still starting up there will always be a point at which `state.members` < whatever constitutes quorum.

Is there perhaps a way to delay the starting of child processes until the cluster `has_quorum?`?

What would be the best way to deal with the startup and quorum?

derekkraan:
Thanks again for the bug report. We can't guarantee that at any point in time, the cluster has quorum. I think the best thing we can do here is add `nil` to the case statement in `handle_call({:start_child, ...})` so that we gracefully handle the case where there is no quorum.

Ivor:
It seems that there is a big difference between lack of quorum because we're still starting up and lack of quorum because we previously had quorum and we've lost some nodes. 

If the starting of processes was something that only started once we atleast reached quorum for the first time then we can prevent this problem. In other words, have the horde start up and hang on witg starting any processes. Monitor the horde in a loop until we have established quorum and then start the processes. 

Is this something you think is feasible?

Ivor:
Alternatively, if we can store the fact that at some earlier time we did have quorum we can consider a failure of quorum to be a legitimate problem, while if we have never had quorum, we can deduce that perhaps we are in this startup phase.

derekkraan:
I'm not sure it's possible to tell the difference between lack of quorum because of start up and lack of quorum because there is a network partition. It could be that we are trying to start the application in the presence of a network partition, for example, in which case we will never reach quorum no matter how long we wait (until the network partition is resolved).

[...]

PR #51, opened 2018-12-07
Record start_time of Supervisor, to allow a grace_period for quorum.

Ivor:
Add the option to set a :grace_period_in_seconds parameter in the config of your project.

```
config :horde, :grace_period_in_seconds
```

In the `init` function we add the time at which the Supervisor is started to the state. 
This allows a period, before `:grace_period_in_seconds` during which not having quorum will be ignored. This allows the normal process of starting the supervisor to succeed. 

After the grace period the normal checks for quorum applies. 

The way I have used this is to start a Task which delays the `start_child` calls to HordeSupervisor by the same grace period. 

```
  def start_horde_children(horde_children) do
    # We're starting the children in a task that is delayed to allow all the
    # members of the cluster to start up.
    Task.start(fn()->
      delay_in_milliseconds = Application.get_env(:horde, :grace_period_in_seconds, 10)*1000
      :timer.sleep(delay_in_milliseconds)
      {:ok, members} = Horde.Cluster.members(MyApp.HordeSupervisor)
      case apply(@distribution_strategy, :has_quorum?, [members]) do
        true ->
          Enum.each(horde_children, fn(child)->
            Horde.Supervisor.start_child(MyApp.HordeSupervisor, child)
          end)
        _no_quorum -> raise "Could not start horde supervisor as quorum was never reached."
      end
    end)
  end
```

A cleaner way would be to perhaps accept the start_child calls but to delay executing them until the grace_period has passed. I am not sure about this though.

### DIFF
--- a/lib/horde/supervisor_impl.ex
+++ b/lib/horde/supervisor_impl.ex
@@ -16,7 +16,8 @@ defmodule Horde.SupervisorImpl do
               processes_updated_counter: 0,
               processes_updated_at: 0,
               shutting_down: false,
-              distribution_strategy: Horde.UniformDistribution
+              distribution_strategy: Horde.UniformDistribution,
+              start_time: nil
   end
 
   def start_link(opts) do
@@ -43,7 +44,8 @@ defmodule Horde.SupervisorImpl do
         node_id: generate_node_id(),
         name: name,
         members_pid: Process.whereis(members_name(name)),
-        processes_pid: Process.whereis(processes_name(name))
+        processes_pid: Process.whereis(processes_name(name)),
+        start_time: DateTime.utc_now
       }
       |> Map.merge(Map.new(Keyword.take(options, [:distribution_strategy])))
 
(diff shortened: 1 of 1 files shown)


Ivor:
If this change is accepted in principle I will update the REAME to reflect this change.

derekkraan:
Hi Ivor,

Thanks for the PR. I can't agree to these changes in principle though.

A better approach to solving this problem would be to simply delay calling `start_child` until the cluster has quorum. No changes are needed to Horde for this. I also won't support changing `start_child` to operate asynchronously (for reasons stated in #49).

Another solution (although I would recommend against it, since simply waiting is simpler and works just as well), is to implement a distribution strategy that has a grace period. This can also be done without altering Horde's internals. In any case, you don't have access to Horde's internal state in the distribution strategy (and I'd like to keep that interface clean, so unless there's a really good reason, I won't be changing that in the future). I am not sure I would accept such a distribution strategy for inclusion in Horde (but you'd be free to use it on your own project, of course, and / or release it as an independent hex package).

Issue #50, opened 2018-12-06
Don't use named processes for dynamic processes


derekkraan:
In (for example) Horde.Supervisor, we start a number of processes that are named. This can create a possible leak in the atom table. Horde should be modified to avoid dynamically creating named processes.

PR #59, opened 2019-01-10
Use sha1 to generate string identifier of a term

farhadi:
When child_id of a process is converted to a binary it is possible that its length exceeds the system limit of atom length so I used SHA1 to always create a fixed size identifier of any child_id.


derekkraan:
It seems like this is related to #50. Do you have any thoughts on that?

farhadi:
Probably we can use Elixir's Registry to keep track of created processes, It will solve both issues.

derekkraan:
I'll merge this, because it is a step in the right direction. I just wanted to bring the other issue to your attention.

PR #62, opened 2019-01-16
WIP: Add transient processes

QuantamHD:


derekkraan:
Hi @QuantamHD,

I don't have time to check this in detail today, but I do think we need to consider how to handle the DynamicSupervisor. IMO it should go something like this:

1. An initial commit with only the original DynamicSupervisor (without any changes at all, not even renaming the module)
2. Let's annotate the DynamicSupervisor with the commit hash from the Elixir github repo, so that we can easily know which patches from elixir core we need to import. (in a comment at the top of the file perhaps?)
3. Then apply any changes that are needed in separate commits.

This is going to make it easier for me to review this PR, and also make it easier in the future to keep us up to date with the Elixir DynamicSupervisor.

Aside from that, thanks for the effort and I'll make it a priority to give you more in depth feedback soon.

Derek

QuantamHD:
Okay only thing I changed here was the name of the module just because if I didn't the modules would conflict and not compile

QuantamHD:
I needed to update the elixir version to 1.7 as the current dynamic supervisor uses the __STACKTRACE__ construct only available in 1.7

derekkraan:
Hi,

I noticed you had taken the `WIP` off the pull request title, but I'm not sure if you're asking me to merge this already?

Cheers
Derek

QuantamHD:
I was wondering if we could create a branch off master to merge this to so that any new changes would be a little easier to see?

derekkraan:
Sounds like a good idea. I've created a branch for that, I'll merge this PR in that branch so you can continue with a fresh PR.

PR #63, opened 2019-01-18
Return `:already_started` when starting an existing child

farhadi:
When starting an already started child, `start_child` must return
`:already_started` with pid of the already started child. To do so:

- name registeration of process supervisor should be done when we start
it, so to prevent from runnning a duplicate one if it's already started.
- When starting an already started child, first we try to start process
supervisor which returns `:already_started` with pid of the already started
process supervisor, in this case we need to retrieve pid of the child
running under this supervisor and return :alreay_started with that pid as the
result of the `start_child`


derekkraan:
I would like to wait on merging this until #62 has progressed further.

derekkraan:
Thanks for the PR. Issues that this PR addresses should be already fixed in master.

PR #64, opened 2019-01-21
Support transient and temporary processes in Horde.Supervisor. Remove Process Canary, Registry and Process Supervisor from Horde.Supervisor

QuantamHD:
This pull request adds support for Horde.Supervisor to supervise a transient process.


derekkraan:
Hi @QuantamHD,

This looks good so far! I haven't looked at this in depth yet but it's encouraging that the testsuite works.

What tests do you propose to add?

There is also an ongoing discussion about whether we should be tracking child_ids in Horde.Supervisor. Do you have any insight to that, having done the work to modify and integrate a custom DynamicSupervisor into Horde?

Derek

QuantamHD:
I think the test would create one horde start a transient process and let it move to completion the join another horde and kill the original horde, and verify that the transient process did not get restarted?

Well, I guess from my perspective the largest piece of work in the DynamicSupervisor code is adding child_id tracking to it. As of today, mainline DynamicSupervisor throws away child_ids I had to add them back so I guess if we stopped using child ids that would make things simpler. My question though is what are the implications for the CRDT what would we use as keys in that case?

lib/horde/horde_dynamic_supervisor.ex
```diff
@@ -897,14 +903,26 @@ defmodule Horde.DynamicSupervisor do
   end
 
   defp maybe_restart_child(_, :normal, pid, _child, state) do
+    if state.process_crdt do
```
QuantamHD: The following three functions are the actual part that makes transient processes work. Everything else in this file is basically just there to make child_ids available in the state.

derekkraan:
It's a tricky question because in DynamicSupervisor you start a process with a child_spec, but all other operations need a pid. This requires the caller to either track the pid themselves, or use a registry, and seems prone to race conditions (what if your process has just died and is in the process of being restarted when you want to terminate it?).

Dropping child ids would also mean that we would be unable to enforce uniqueness in the Supervisor. Enforcing uniqueness would then require the use of a Registry as in the regular DynamicSupervisor. This becomes tricky when using Horde.Registry, because it is eventually consistent, so while the normal Registry will return an error immediately if your process is not unique, allowing you to kill it, Horde.Registry doesn't have this capacity. Horde.Registry also cannot kill your process after the fact (that is really a Supervisor responsibility) so it also has no way to correct the problem later, when states have synchronized.

Ideally I would like to keep the child_id, but it also feels very strange that we are not compatible with Elixir's DynamicSupervisor in this way, and there's also the question of whether this will make it more difficult for us to stay in sync with Elixir's DynamicSupervisor in the future (although I'm not expecting them to make many changes going forward).

In terms of the test, I think we should be testing both scenarios (normal exit & abnormal exit) for both temporary and transient processes. It should be sufficient to see that the process has been removed from the supervisor's definition of which processes it needs to supervise. (You can cheat and inspect the state of the SupervisorImpl to do this), no need IMO to join a new Horde and kill the original one.

QuantamHD:
In my opinion, we have a very different use case than Elixir.DynamicSupervisor and it would be okay to break compatibility in this way. I think that it's a nice to have a similar interface that has fewer bugs than an exact match with more limitations and potential race conditions.

QuantamHD:
Also, I'm not sure that we should allow temporary processes in Horde given our earlier conversations about how they really can't be restarted.

derekkraan:
I actually can't really think of a reason why DynamicSupervisor would support temporary processes, other than that Supervisor already does. I think Supervisor supports it so that you can take advantage of the startup boilerplate / behaviour. DynamicSupervisor doesn't have this though.

So let's skip temporary processes for now. I'm not opposed out of principle to adding them at a later date, but let's stick with YAGNI and wait for someone to come with a good argument for them.

derekkraan:
Ok, I now understand the reasoning behind why DynamicSupervisor supports temporary processes. It's potentially useful for ensuring that your process exits when the supervisor exits. This raises some weird questions about how this should work in a distributed context though, so I'm still ok with not implementing it.

derekkraan:
Actually, thinking about this further, perhaps we should implement `temporary` processes after all.

Consider for example a web crawler: we might be using Horde.Supervisor to distribute the load over a cluster. The process could fail for any number of reasons, but we don't want to restart, just report back if we have found something. I could see this being a realistic scenario. I also think the proper behaviour is to never restart them (but still shut them down if the node is being shut down).

What are your thoughts?

QuantamHD:
Okay, so I ended up adding temporary children and removed the process canary graceful shutdown structure and added graceful shutdown behaviors directly in Horde.DynamicSupervisor.

derekkraan:
Hi @QuantamHD, you closed this PR and deleted your branch, are you planning on opening a new one?

QuantamHD:
I was informed by my company that I needed to fill out some paper work before contributing any further. It will take a few days to process though.

derekkraan:
Thanks for the update @QuantamHD

QuantamHD:
Okay should be all good I had to add NOTICES to align with Apache v2

lib/horde/horde_dynamic_supervisor.ex
```diff
@@ -881,9 +1104,28 @@ defmodule Horde.DynamicSupervisor do
...
+        shutdown: shutdown,
+        type: type,
+        modules: modules
+      }
+
+      GenServer.cast(state.graceful_shutdown_manager, {:shut_down, child_spec})
+    end
+  end
+
+  defp handle_graceful_shutdown_horde(_, _) do
```
derekkraan: what other cases are we handling that this will match?

lib/horde/horde_dynamic_supervisor.ex
```diff
@@ -914,12 +1159,26 @@ defmodule Horde.DynamicSupervisor do
   end
 
   defp maybe_restart_child(:temporary, reason, pid, child, state) do
+    remove_child_from_horde(state, pid)
     report_error(:child_terminated, reason, pid, child, state)
     {:ok, delete_child(pid, state)}
   end
 
+  defp remove_child_from_horde(state, pid) do
+    if state.root_name do
```
derekkraan: If we check for this in `init`, then we can remove this `if`.

lib/horde/supervisor.ex
```diff
@@ -65,12 +65,6 @@ defmodule Horde.Supervisor do
   """
   def start_child(supervisor, child_spec) do
     child_spec = Supervisor.child_spec(child_spec, [])
-
-    if :transient == Map.get(child_spec, :restart, nil) do
```
derekkraan: :clap:

lib/horde/supervisor_impl.ex
```diff
@@ -140,17 +140,7 @@ defmodule Horde.SupervisorImpl do
...
-            {:undefined, :restarting, _, _} = child ->
-              child
-
-            {:undefined, pid, _, _} ->
-              Enum.filter(Supervisor.which_children(pid), fn
-                {Horde.ProcessCanary, _, _, _} -> false
-                _ -> true
-              end)
-          end)
+          Horde.DynamicSupervisor.which_children(supervisor_pid)
```
derekkraan: :heart_eyes:

lib/horde/supervisor_impl.ex
```diff
@@ -167,21 +157,7 @@ defmodule Horde.SupervisorImpl do
...
-              count = Supervisor.count_children(pid)
-
-              %{
-                count
-                | workers: count.workers - 1,
-                  active: count.active - 1,
-                  specs: count.specs - 1
-              }
-          end)
+          Horde.DynamicSupervisor.count_children(supervisor_pid)
```
derekkraan: :heart_eyes:

lib/horde/supervisor_impl.ex
```diff
@@ -345,11 +327,20 @@ defmodule Horde.SupervisorImpl do
...
-      :ok =
-        DynamicSupervisor.terminate_child(
+      pid =
+        Horde.DynamicSupervisor.find_pid(
           processes_supervisor_name(new_state.name),
-          get_child_pid(new_state, id)
+          id
         )
+
+      if pid do
```
derekkraan: Ok, so here we are checking whether the pid exists, but on line 92 we do not. Is there a race condition here that I'm not aware of?

My thinking is: the DynamicSupervisor will *always* have a pid for every process under its supervision (if it could not start a child process after X retries within Y minutes, the supervisor itself will die).

Actually come to think of it, the following could happen:
1. we call `Horde.DynamicSupervisor.find_pid`
2. the process belonging to the pid that is returned dies
3. Horde.DynamicSupervisor restarts the process (now has a new pid, and doesn't know about our pid anymore)
4. We call Horde.DynamicSupervisor.terminate_child

This indicates to me that we need to bundle these two calls into one: a terminate_child that accepts an `id` instead of a `pid`.

derekkraan: And if we have this, then we can remove all mention of child pids from this module (should save some bookkeeping / code).

lib/horde/supervisor_impl.ex
```diff
@@ -218,9 +194,15 @@ defmodule Horde.SupervisorImpl do
     {:noreply, mark_alive(state, true)}
   end
 
-  defp get_child_pid(%{name: name}, child_id) do
-    [{pid, _}] = Registry.lookup(registry_name(name), child_id)
-    pid
+  def handle_cast({:remove_process_tracking, child_id}, state) do
```
derekkraan: Perhaps we can come up with a better name for this than `remove_process_tracking`. Something that indicates its intent (ie, the supervisor has indicated that the child has terminated and should not be restarted).

lib/horde/supervisor_impl.ex
```diff
@@ -509,36 +501,10 @@ defmodule Horde.SupervisorImpl do
...
-                processes_supervisor_name(state.name),
-                process_supervisor_pid
-              )
-
-              {:error, error}
-          end
+    Enum.map(children, fn child ->
+      case Horde.DynamicSupervisor.start_child(processes_supervisor_name(state.name), child) do
+        {:ok, process_pid} ->
+          {{:ok, process_pid}, child}
```
derekkraan: :heart_eyes:

lib/horde/horde_dynamic_supervisor.ex
QuantamHD: On line 859 it protects against the child spec being nil for a given pid.

lib/horde/supervisor_impl.ex
QuantamHD: maybe aborted_child?

lib/horde/supervisor_impl.ex
QuantamHD: So with the old code, we had an implied concept of local ownership(Registry) and global ownership(CRDT). With Process Canary removed we lost the concept of local ownership so what can happen here is when the Supervisor tries to kill it's not owned children all it has left is the global data which is a race i.e. often what would happen is that it would try to remove a pid that didn't belong to this supervisor. So what I had to do was give the local supervisor the ability to map child_ids to pids. I'm checking here just to make sure that the pid exists and is not nil maybe unnecessarily. The reason I don't check in 98 is that method affects global data whereas this one is pure local data.

QuantamHD: Ah I think I see your point this can't be two different calls otherwise the race will occur.

lib/horde/supervisor_impl.ex
QuantamHD: Or another would be forget_child.

QuantamHD: What do you think?

lib/horde/horde_dynamic_supervisor.ex
derekkraan: I am still wondering about this line. I am also wondering whether line 1071 should be removed (and whether this will solve the issue of the missing child specs).

The way DynamicSupervisor terminates a pid is by monitoring it, sending a `:shutdown` signal, and then processing the `:DOWN` tuples as they come in. I guess a `:DOWN` tuple could come out of order (eg, spuriously), which is why there is still a catch-all match on 1070. But as far as I can tell, we should *always* come out at either 1065 or 1075 for every pid that we are terminating.

Thoughts?

derekkraan:
This is looking great, there is just the one unresolved issue that I'd like to discuss before merging.

lib/horde/horde_dynamic_supervisor.ex
QuantamHD: After going through the code and tracing it in my head I don't think there is a case for it. I removed it, I tend to like to add catch all things to avoid crashing the supervisor, but if something like that happens "let it crash".

QuantamHD:
Okay, everything should be good now.

derekkraan:
Thanks @QuantamHD :heart: !!

QuantamHD:
No problem, have a good one Derek.

PR #73, opened 2019-02-27
replace add_neighbours with set_neighbours

dmytronasyrov:


derekkraan:
Hi @dmytronasyrov, please see #72, which already addresses this issue (but needs a little more work before it can be merged).

PR #95, opened 2019-04-16
An exception is raised when `Horde.UniformDistribution.choose_node/2` and no alive members are available

flaviogrossi:
If all cluster members are in `:dead` or `:shutting_down` state, when `choose_node/2` is called an exception is raised because a division by zero is attempted.
This often happens in tests, because of the fast termination of cluster members (it happens in horde's supervisor tests too).

Here i added a return value of `{:error, :no_alive_nodes}` when no nodes are available, similar to what happens in the uniform quorum distribution case.

A test is added to expose the bug.

### DIFF
--- a/lib/horde/uniform_distribution.ex
+++ b/lib/horde/uniform_distribution.ex
@@ -14,9 +14,14 @@ defmodule Horde.UniformDistribution do
       end)
       |> Enum.sort_by(fn %{name: name} -> name end)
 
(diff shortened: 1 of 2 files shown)


derekkraan:
Hi, thanks for the PR! Can you remove your .swp file from the diff and squash this into one commit? Then I will merge it!

flaviogrossi:
Yeah, sorry about that, I noticed that seconds after the PR. Should be all ok now.

derekkraan:
:+1::+1::+1:

Issue #54, opened 2018-12-20
Horde.Supervisor.terminate_child/2 non consistent with DynamicSupervisor 


xadhoom:
According to Horde docs and function spec, terminate_child/2 should use the child pid as child_id to terminate, like DynamicSupervisor does.

But does not work, and looking into tests seems that we need to use the id of the process (which works), which is (maybe) fine, but should at least be documented.

derekkraan:
Thanks for the bug report. I would gladly accept a pull request to fix this issue. Looking at the documentation, the description is correct but the typespec does specify a `pid`.

xadhoom:
mmmh, not really :)
The doc states that `Works like DynamicSupervisor.terminate_child`, which accepts a pid.

Horde.Supervisor wants an id instead. So I'll be happy to contribute, but I need to understand what is really wanted.

If an `id` is the wanted params, specs & docs must be fixed (glad to do it).
Otherwise, if we want API parity, `supervisor_impl.ex` must be fixed (easy to do, also).
As a third choice, we may support both :) (maybe is the better choice ?)

Let me know what is your idea and I'll try to do something.

derekkraan:
Ah yes, that's true.

Ideally we would have API parity with Elixir's DynamicSupervisor. Right now we don't have that. The main difference is that Horde.Supervisor knows about ids and actually uses them internally, while DynamicSupervisor does not (it ignores child ids).

The more fundamental difference between Horde.Supervisor and DynamicSupervisor is that Horde.Supervisor has a notion of process identity, and DynamicSupervisor does not.

Here's a concrete example:

```
Horde.Supervisor.start_child(supervisor, %{id: "foo", start: {...}})
#=> {:ok, pid}
Horde.Supervisor.start_child(supervisor, %{id: "foo", start: {...}})
#=> {:error, {:already_started, pid}}

DynamicSupervisor.start_child(supervisor, %{id: "foo", start: {...}})
#=> {:ok, pid}
DynamicSupervisor.start_child(supervisor, %{id: "foo", start: {...}})
#=> {:ok, pid}
```

[...]

PR #104, opened 2019-05-03
save child pid into state in order to return the correct pid on already_started error

### LINKED ISSUE #102: Horde.Supervisor.start_child returns nil pid when :already_started
When starting a new child with `Horde.Supervisor.start_child/2` and the child is already started, the error `{:error, {:already_started, pid}}` is returned, but pid is always nil.

Having the pid in the returned error is commonly used to perform operation on the process, for example try to start the process -> already started -> GenServer.call(pid, whatever).

xadhoom:
this should fix #102

### DIFF
--- a/test/supervisor_test.exs
+++ b/test/supervisor_test.exs
@@ -75,16 +75,18 @@ defmodule SupervisorTest do
 
   describe ".start_child/2" do
     test "starts a process", context do
-      assert {:ok, _pid} = Horde.Supervisor.start_child(context.horde_1, context.task_def)
+      assert {:ok, pid} = Horde.Supervisor.start_child(context.horde_1, context.task_def)
 
-      assert {:error, {:already_started, _child}} =
-               Horde.Supervisor.start_child(context.horde_1, context.task_def)
+      assert_receive {:process_started, ^pid}
     end
 
     test "returns error when starting same process twice", context do
       assert {:ok, pid} = Horde.Supervisor.start_child(context.horde_1, context.task_def)
 
-      assert_receive {:process_started, ^pid}
+      assert is_pid(pid)
+
+      assert {:error, {:already_started, ^pid}} =
+               Horde.Supervisor.start_child(context.horde_1, context.task_def)
     end
 
     test "starts a process with id that doesn't implement String.Chars", context do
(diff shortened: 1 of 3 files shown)


test/supervisor_test.exs
```diff
@@ -75,16 +75,19 @@ defmodule SupervisorTest do
...
     end
 
     test "returns error when starting same process twice", context do
       assert {:ok, pid} = Horde.Supervisor.start_child(context.horde_1, context.task_def)
 
-      assert_receive {:process_started, ^pid}
+      assert {:error, {:already_started, child}} =
+               Horde.Supervisor.start_child(context.horde_1, context.task_def)
+
+      assert is_pid(child)
```
derekkraan: you can clean up this part by asserting on line 86 as follows:

`assert {:error, {:already_started, ^pid}} = ...`

derekkraan:
Hi @xadhoom, first of all, thanks for the PR. I'm going to think about this one for a couple more days before I come back with comments.

Cheers
Derek

xadhoom:
Mmmh yes, but no one assures me that pid is really a pid :). What if both
returns nil? I'll clean it up but add a type assertion if ok for you.

Il lun 6 mag 2019, 08:56 Derek Kraan <notifications@github.com> ha scritto:

test/supervisor_test.exs
xadhoom: Mmmh yes, but no one assures me that pid is really a pid :). What if both returns nil? I'll clean it up but add a type assertion if ok for you.

xadhoom:
ok force pushed an update and squashed all commits into one, since is a single feat

derekkraan:
Hi,

First of all, thanks for the PR, and I want to say that the effort is appreciated. I've thought about this, and I think we should support this another way. We should query the relevant Horde instance to ask it about the pid of the process if we need the pid for this error tuple.

Tracking the pids will potentially cause a LOT of churn in the CRDT, to solve what is essentially a race condition, and I want to avoid this.

Will you make the necessary changes to this PR?

Cheers,
Derek

xadhoom:
Don't understand what you mean with "solve what is essentially a race condition"... there're a lot of common cases where you want to start a process in a idempotent way (for example start user/device/whatever tracker when receiving a specific event), and I don't care if started already or not, but I want to call the pid and maybe update it with the new informations.

Also if OTP provides the pid to such APIs, there's a lot of sense in providing it :) (I assume who designed OTP is much smarter and have seen many more use cases than me)

Probably the same can be accomplished using a registry, but is more convoluted (ask registry, start if not there, what if 2 events are received at almost same time? you'll need to handle already_started error anyway)

About CRDT churn... why? It may be a problem if your process dies badly and get restarted often, which should not be the normal running condition. In any other case the data is put in the CRDT at creation, so not creating additional CRDT work (same for deletion). Or I'm missing something?

Said that, the PR can be updated, but I need some time to figure out how to find the owning Horde instance and add proper APIs (this will be indeed slower than keeping a copy of the information on local state like all other data without having to to a remote call to another node)

What do you think?

derekkraan:
Hi,

Just a quick note, I'm still thinking about this PR, just not sure yet what to do. I want to ge it right. Just so you don't think I'm ignoring this PR. :stuck_out_tongue: 

Derek

derekkraan:
Specifically I'm thinking about our conversation here: https://github.com/derekkraan/horde/issues/54 and whether it's maybe a better idea to ignore `id` (we'd still need an internal `id` of course, but we could generate that). This would bring us into better alignment with `DynamicSupervisor` and also solve the issue of this PR.

derekkraan:
If we move to ignoring `id`, like DynamicSupervisor does, then we have to track the `pid` anyways, otherwise we can't get things like `terminate_child` to work :man_facepalming:

derekkraan:
I'm going to merge this into another branch and do some experiments on it there, see if we can become more like DynamicSupervisor on the whole.

Issue #111, opened 2019-05-07
Getting started documentation may not be valid anymore?


pdgonzalez872:
Env details:

```
# OS -> OSX
Erlang/OTP 21 [erts-10.0.5] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

Interactive Elixir (1.7.2) - press Ctrl+C to exit (type h() ENTER for help)
```

Issue:
I wanted to add a `GenServer` to the `Horde.Supervisor` supervision tree but was not able to.

Expected behavior:
Can add a child to the supervision tree correctly. Would display in the observer.

Sample app that displays the behavior. You can check the branch out and run `mix test`:
https://github.com/pdgonzalez872/horde/tree/add-horde-only-example/examples/horde_example

I added a silly assertion to fail first before writing a passing test, but I got a different error. I tried in `iex` and got the same behavior. The test will fail with this error:
```
horde_example [add-horde-only-example] :> mix test
Compiling 1 file (.ex)

15:10:32.721 [info]  Starting Horde.SupervisorImpl with name HordeExample.DistributedSupervisor


  1) test use this as an example should work (NetsplitTest)
     test/acceptance_test.exs:5
     ** (MatchError) no match of right hand side value: {:error, {:undef, [{GenServer, :start_link, [], []}, {Horde.DynamicSupervisor, :start_child, 3, [file: 'lib/horde/horde_dynamic_supervisor.ex', line: 765]}, {Horde.DynamicSupervisor, :handle_start_child, 2, [file: 'lib/horde/horde_dynamic_supervisor.ex', line: 751]}, {:gen_server, :try_handle_call, 4, [file: 'gen_server.erl', line: 661]}, {:gen_server, :handle_msg, 6, [file: 'gen_server.erl', line: 690]}, {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 249]}]}}
     code: :ha = HordeExample.add_child_to_supervisor()
     stacktrace:
       test/acceptance_test.exs:6: (test)



Finished in 0.08 seconds
1 test, 1 failure
```

Potential solutions:
- Are the docs here still applicable? https://github.com/derekkraan/horde#usage  This is the line I question: `Horde.Supervisor.start_child(MyApp.DistributedSupervisor, %{id: :gen_server, start: {GenServer, :start_link, []}})`. Maybe there was a change and the docs have not been updated yet.
- I missed a step somewhere.

[...]

PR #112, opened 2019-05-09
Update readme with examples

pdgonzalez872:
Should close https://github.com/derekkraan/horde/issues/111

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,17 @@ end
 Adding a child to the supervisor:
 
 ```elixir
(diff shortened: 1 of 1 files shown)


pdgonzalez872:
Is this any better @derekkraan, I'm unsure. I don't want to overcomplicate things. Please let me know. Thank you.

derekkraan:
Hi, thanks for the effort! This is an improvement on what was there. I'll be taking a pass at the documentation soon, because I do think it could be a lot better.

PR #114, opened 2019-05-16
fix pattern match

davec82:
Sometimes when horde nodes are connected  there is a MatchError in delete_child fn


derekkraan:
Thanks for the PR! I noticed a couple of other places where this can also happen, so I'll fix those now as well.

derekkraan:
No additional fixes necessary, this was the only one. :+1:

PR #115, opened 2019-05-16
Registry: Prevent local duplicate registration

arjan:
This patch makes Registry.register/3 more in line with its Elixir
equivalent. In fact, I copied the test cases from Elixir's Registry.

Duplicate registrations can still happen (overwrite each other) within
members of the cluster. This will be dealt with in another PR


derekkraan:
Thanks! :raised_hands:

PR #118, opened 2019-05-16
 Send :name_conflict message to losing process when registries merge

arjan:
When 2 clustered registries register the same name at exactly the  same time, it will seem like name registration succeeds for both  registries. The function returns `{:ok, pid}` for both of these  calls.

However, due to the eventual consistent nature of the CRDT, a  conflict resolution will take place, and the CRDT will pick one of  the two processes as the "winner" of the name. The losing process   will then receive a message to indicate that it has lost the name  registration. The message looks like this:

    {:name_conflict, {name, value}, registry_name, winning_pid}

### DIFF
--- a/lib/horde/registry.ex
+++ b/lib/horde/registry.ex
@@ -119,8 +119,8 @@ defmodule Horde.Registry do
   However, due to the eventual consistent nature of the CRDT, a
   conflict resolution will take place, and the CRDT will pick one of
   the two processes as the "winner" of the name. The losing process
-  will then receive a message to indicate that it has lost the name
-  registration. The message looks like this:
+  will be sent an exit signal (using `Process.exit/2`) with the
+  following reason:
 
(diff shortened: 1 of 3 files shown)


derekkraan:
Thanks this looks great!

I was thinking about this last night, maybe it would be better to send an exit signal with the reason. Then it's still a message that processes can respond to, but the default behaviour (exiting vs ignoring) is perhaps better.

What do you think?

derekkraan:
Consider: Elixir.Registry (and Horde.Registry) links your process to itself, so if the registry goes down, you get an exit signal. Also, it's common to write `{:ok, pid} = Horde.Registry.register(...)` in the `init/1` callback of your process, and also just let it crash if that fails. (At least, this is what I do).

arjan:
Yes so to catch this signal you then need to be trapping exits. I think that's fine. I'll fix this, hold on

derekkraan:
Yes, this is looking great. Thanks!

PR #119, opened 2019-05-17
Fix flaw where name registration was lost entirely on conflict

arjan:
The exiting of the loser process caused the name registration to be removed from the CRDT entirely.

### DIFF
--- a/lib/horde/registry_impl.ex
+++ b/lib/horde/registry_impl.ex
@@ -160,16 +160,13 @@ defmodule Horde.RegistryImpl do
   defp process_diff(state, {:add, {:key, key}, {member, pid, value}}) do
     link_local_pid(pid)
 
-    case :ets.lookup(state.pids_ets_table, pid) do
-      [] -> :ets.insert(state.pids_ets_table, {pid, [key]})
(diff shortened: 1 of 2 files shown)


arjan:
Not sure that these failures are caused by my fix..

derekkraan:
probably not. the tests are a little sensitive to time, not really sure how we could fix that, as we are testing an eventually consistent system ... :thinking:

arjan:
wait I have a better solution than this, don't merge

derekkraan:
I was about to write that I think this can be solved in a better way, but I'll wait to see what you come up with now ;)

arjan:
`Process.unlink`

derekkraan:
Ah, that's not what I had in mind. I think we need to clean up the ets tables properly when we detect that a pid loses. So in the `process_diff` function.

arjan:
Yes, coming to realise that as well now

arjan:
This should be better

derekkraan:
I think we're really close. But we shouldn't delete the whole entry, just remove the key from its list of keys.

arjan:
True

arjan:
this patch should do the trick, I think

derekkraan:
:ok_hand:

arjan:
Awesome

Issue #82, opened 2019-03-29
Documentation confusion between DynamicSupervisor and Supervisor API


Shikanime:
**Describe the bug**

Hi,

The documentation has a lot of reference with the DynamicSupervisor but seems to act as a Supervisor which is a little confusing.
And the return seem to be inconsistent by returning nil pid  on already_started.

**To Reproduce**

```elixir
defmodule MyApp.Supervisor do
  # Automatically defines child_spec/1
  use Supervisor

  def start_link(init_arg) do
    Supervisor.start_link(__MODULE__, init_arg)
  end

  @impl true
  def init(_init_arg) do
    children = [
      {MyServer, []}
    ]

    Supervisor.init(children, strategy: :one_for_one)
  end
end

defmodule MyServer do
  use GenServer

  def start_link(state) do
    GenServer.start_link(__MODULE__, state)
  end

  ## Callbacks

  @impl true
  def init(stack) do
    {:ok, stack}
  end
end

{:ok, _} = DynamicSupervisor.start_link(name: MyApp.CusterDynamicSupervisor, strategy: :one_for_one)

DynamicSupervisor.start_child(MyApp.CusterDynamicSupervisor, {MyApp.Supervisor, []})
# {:ok, #PID<0.419.0>}
DynamicSupervisor.start_child(MyApp.CusterDynamicSupervisor, {MyApp.Supervisor, []})
# {:ok, #PID<0.422.0>}

DynamicSupervisor.which_children(MyApp.CusterDynamicSupervisor)
# [
#   {:undefined, #PID<0.419.0>, :supervisor, [MyApp.Supervisor]},
#   {:undefined, #PID<0.422.0>, :supervisor, [MyApp.Supervisor]}
# ]

{:ok, _} = Supervisor.start_link([], name: MyApp.CusterSupervisor, strategy: :one_for_one)

# Supervisor.start_child(MyApp.Horde, {MyApp.CusterSupervisor, []})
# {:ok, #PID<0.419.0>}
# Supervisor.start_child(MyApp.Horde, {MyApp.CusterSupervisor, []})
# {:error, {:already_started, #PID<0.419.0>}}

# Supervisor.which_children(MyApp.Horde)
# [{MyApp.Supervisor, #PID<0.419.0>, :supervisor, [MyApp.Supervisor]}]

{:ok, _} = Horde.Supervisor.start_link(name: MyApp.CusterDistributedSupervisor, strategy: :one_for_one)

[...]

PR #120, opened 2019-05-19
Adds example for dynamic process creation

pkinney:
A simple example of creating processes on demand with Horde.  Hopefully this simplifies gettings started with Horde for some folks looking to use it for this type of use-case.

This was also my way of trying to sort through the current behaviors discussed in a couple of issues (#104, #82, and #22). These all seem to boil down to needing an idempotent way of starting a new process and the ability to call that process immediately after creation.  I have an implementation in the `Counter.Router` module, but this may not be the best way to do it. I'm open to suggestions.

From the [README](https://github.com/pkinney/horde/blob/counter-example/examples/counter/README.md):

> Internally, `Router` uses `Horde.Registry.lookup` to determine if a `Worker` process for a given `name` exists and if not starts one.  Then it either calls that process using the PID returned or using the `:via` method as outlined in the Horde documentation.


pkinney:
Added handling of the PID being returned from `Horde.Supervisor.start_child` as is being done in PR #122. Pulling those changes form that PR in and testing removes the race condition seen in some situations where the Registry is not yet consistent across nodes, but the Supervisor knows that the process is already started.

derekkraan:
Hi @pkinney,

I have been thinking about what to do about this PR for a while now (sorry about the delay). I think you are on to something here, there is no good documentation on how to dynamically start processes.

But, I am also hesitant to merge this PR as-is. I have a hard time keeping the existing small example app up to date and working, and I don't think I want to add another example app to the mix (at least until I'm sure the API has stabilized, aka after v1.0.0).

My suggestion would be to consider contributing to the documentation of Horde by writing a guide. I have just added a couple, so it should be easy to follow that example (start by looking in the mix.exs file). I think this would have a lot more impact than an example app.

Cheers,
Derek

pkinney:
@derekkraan: That makes sense to me.  My primary goal with putting the example together to make sure that the way it handles this use case is in line with Horde's vision.  If so, then I think putting this approach in a guide would serve the same purpose for the others that might need it.

I'll take a deeper look at what's there now and see where I can fit it in.  Thanks!

PR #121, opened 2019-05-20
fix Registry specs

flaviogrossi:
this fixes the typespecs to avoid dialyzer errors in projects using this library.


derekkraan:
Thanks!!

PR #122, opened 2019-05-21
More compatible with DynamicSupervisor

derekkraan:
This PR increases compatability with DynamicSupervisor.

Changes:
- `start_child/2` ignores any `id` in the child_sped argument, and generates its own id for internal use.
- `terminate_child/2` accepts a pid as second argument instead of an id
- `which_children/1` now returns `{:undefined, pid, type, modules}`

This PR should solve the following issues: #82, #54.

This PR also means `Horde.Supervisor` no longer dedups processes across a cluster. Just like `Elixir.DynamicSupervisor`, you're expected to use `Horde.Registry` to gain this functionality.


arjan:
Nice change! :+1: 

Very good to get things more API-compatible with DynamicSupervisor.

PR #123, opened 2019-05-21
[POC] Name conflicts don't count towards the max restart intensity

### LINKED ISSUE #116: Registry drops all entries when second node joins
I've created a sample app with instructions on how to replicate https://github.com/uberbrodt/test_horde. 

Briefly, I'm starting a Horde.Supervisor and a Horde.Registry, and a couple of processes that:
1. Join any connected nodes via Horde.Cluster.set_members/2
2. execute a callback to start 10 processes under the Horde.Supervisor, registering them with a via-tuple

As stated in the title, the registry entries seem to work when the first node starts, but when the second one starts and joins, all entries are wiped out for both nodes. Interestingly, all supervised processes move from the first node to the second node. I expected that `UniformDistribution` would place half on the first node and half on the second.

I'm testing with elixir 1.8.1 and OTP 21.3.6. Thanks in advance for looking into this.

derekkraan:
These name conflicts are not real failures, so we don't want them to
trigger max restart intensity. In the event of a netsplit that is
healing, there may be a LOT of these happening, so it's almost
guaranteed that we would exceed the max restart intensity.

Should fix #116 

Note: I'm not sure this is actually a good idea, since we are letting the supervisor know about a detail of the registry.

~This PR also needs a test.~

### DIFF
--- a/lib/horde/horde_dynamic_supervisor.ex
+++ b/lib/horde/horde_dynamic_supervisor.ex
@@ -994,6 +994,10 @@ defmodule Horde.DynamicSupervisor do
     end
   end
 
+  defp maybe_restart_child(_, {:name_conflict, _, _, _}, pid, child, state) do
+    restart_child_no_max_intensity(pid, child, state)
+  end
+
   defp maybe_restart_child(:permanent, reason, pid, child, state) do
     report_error(:child_terminated, reason, pid, child, state)
     restart_child(pid, child, state)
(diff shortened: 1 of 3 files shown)


xadhoom:
Well, standard supervisors simply don't have such mechanism and often you have to deal with it.
But such handling is application specific, and often you have to add some "delay" between restarts (see for example https://github.com/klarna/supervisor3) and always have an exit way to avoid infinite loops.

Maybe can be optional and with some tuning parameters, with reasonable defaults?

jm2c

derekkraan:
@xadhoom This will only trigger once per detected conflict.

The way this will work in applications using Horde today is:
1. a conflict is detected, the process gets an exit signal `{:name_conflict, _, _, _}`
2. the process dies, and is restarted (no impact to max restart intensity)
3. the restarted process tries to register itself, this fails and the process dies (max restart intensity impacted)
4. if there are enough processes then the supervisor will also die.

So this PR doesn't actually change the default behaviour much (it slightly delays the max restart intensity from triggering).

People have two options:
1. trap exits and handle this exit signal (this was already an option actually).
2. return `:ignore` from `start_link` when `{:error, {:already_registered, pid}}` is returned (which will remove the process from the supervisor).

Actually now I am thinking about it, perhaps the best option is to just expecte people to handle the exit signal in whatever way they wish (eg: by shutting down the process gracefully). Then we can kill this PR and just add some documentation.

derekkraan:
Yep, I think it's reasonable to expect people to work around this, and I will add some documentation highlighting the `{:name_conflict, _, _, _}` exit signal that's new.

PR #125, opened 2019-05-30
Fix Hore.Supervisor.start_child/2 to handle :ignore from child

arjan:


arjan:
Not sure if the CI fail has something to do with my patch :thinking:

derekkraan:
Just timing issues.

PR #127, opened 2019-06-02
Check if process alive on `Horde.Registry.register/3`.

derekkraan:
This will solve a race condition caused by a registry not being aware
that a remote process has died (eg, during node shut down)


arjan:
Nice, this looks good, and works as expected :-D

derekkraan:
👍 thanks for testing this

PR #128, opened 2019-06-03
Hash the child spec minus the id for consistent hashing.

derekkraan:
With #122 we started assigning random ids to bring us in line with DynamicSupervisor's API. This broke consistent hashing. This PR restores that, by hashing the child spec minus the id.

Caveat: if you are starting processes that only differ by the id, and want them to end up on different nodes, then you will have to add something to the child_spec to differentiate them, or alternatively supply your own distribution_strategy.

Also removes xxhash as dependency.

_This PR needs a test._

### DIFF
--- a/lib/horde/supervisor_impl.ex
+++ b/lib/horde/supervisor_impl.ex
@@ -141,8 +141,10 @@ defmodule Horde.SupervisorImpl do
 
     child_spec = Map.put(child_spec, :id, :rand.uniform(@big_number))
 
+    distribution_id = :erlang.phash2(Map.drop(child_spec, [:id]))
+
     case state.distribution_strategy.choose_node(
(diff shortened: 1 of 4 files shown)


redrabbit:
> Caveat: if you are starting processes that only differ by the id, and want them to end up on different nodes, then you will have to add something to the child_spec to differentiate them, or alternatively supply your own distribution_strategy.

Hi @derekkraan, the changes of this PR make it hard to implement a custom distribution strategy based on the child_spec given to `Horde.DynamicSupervisor.start_child/2`.  Is `Horde.DynamicSupervisor` meant to be used only with a consistent hash mechanism?

In my project I need to choose a node based on some criteria given to the child_spec. Is there any way to achieve this?

PR #135, opened 2019-07-09
Make Horde.Supervisor child_spec overridable

### LINKED ISSUE #107: Improve module based supervisor
Current module based supervisor implementation does not allow setting supervisor options like max restarts, timers and so on, that's because of:

* child_spec is not overridable
* init/1 in standard module-based supervisor is expected to return a Supervisor.init call, in horde returns some options that's not clear what are used for

My suggestion is to make the module-based supervisor as the standard one. If is too complicated maybe make child_spec overridable in the macro?

ordepdev:
Fixes #107

### DIFF
--- a/test/supervisor_test.exs
+++ b/test/supervisor_test.exs
@@ -89,6 +89,28 @@ defmodule SupervisorTest do
       {:ok, members} = Horde.Cluster.members(:init_sup_test_1)
       assert 2 = Enum.count(members)
     end
+
+    test "can use `child_spec` function to override defaults from Horde.Supervisor" do
+      spec = %{
+        id: 123,
+        start:
+          {TestSupervisor3, :start_link,
+           [
+             custom_id: 123,
+             name: :init_sup_test_3,
+             strategy: :one_for_one
+           ]},
+        restart: :transient,
+        type: :supervisor
+      }
+
+      assert spec =
+               TestSupervisor3.child_spec(
+                 custom_id: 123,
+                 name: :init_sup_test_3,
+                 strategy: :one_for_one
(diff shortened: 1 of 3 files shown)


ordepdev:
Hmm, running the tests locally it passes with:
```
Finished in 37.9 seconds
3 properties, 70 tests, 0 failures, 1 skipped
```

Still, is failing in circleci with:

```
1) test .set_members/2 can join and unjoin registry with set_members (ClusterTest)
     test/cluster_test.exs:164
     match (=) failed
     code:  assert 1 = Enum.count(members)
     right: 2
     stacktrace:
       test/cluster_test.exs:190: (test)

Finished in 37.6 seconds
3 properties, 70 tests, 1 failure, 1 skipped
```

ordepdev:
Oh, it was my mistake 😅

derekkraan:
Also a good chance it was a wonky test! There are a couple of those here.

test/supervisor_test.exs
```diff
@@ -89,6 +89,28 @@ defmodule SupervisorTest do
...
+           [
+             custom_id: 123,
+             name: :init_sup_test_3,
+             strategy: :one_for_one
+           ]},
+        restart: :transient,
+        type: :supervisor
+      }
+
+      assert spec =
```
derekkraan: This assertion isn't asserting anything I don't think. Should be `^spec` I think.

ordepdev: Ohhh, it was testing... nothing  🙈 Already fixed!

derekkraan:
Looks great, thanks!!

PR #137, opened 2019-07-17
Update libcluster guide to swop order of members to {name, node}

Ivor:


derekkraan:
Looking good, thanks! Can you also fix it in the static membership section of the guide?

Ivor:
Updated! Thanks for the awesome library and the quick response.

derekkraan:
Thanks for the PR!

PR #138, opened 2019-07-19
Message of set_members documented

F2011B:
I tried to implement a DeltaCrdt handoff using the Tanx example from 2018, 
I was not able to understand where the message :set_members originated. 
Maybe it is already documented in hex, however I have not found it on the first glance.

### DIFF
--- a/lib/horde/cluster.ex
+++ b/lib/horde/cluster.ex
@@ -3,6 +3,8 @@ defmodule Horde.Cluster do
 
   @moduledoc """
   Public functions to join and leave hordes.
+  It will also send a :set_members call to every member of the horde Cluster,
+  with content members [{member_one_name, member_one_node}, {member...}]
 
   Calling `Horde.Cluster.set_members/2` will join the given members in a cluster. Cluster membership is propagated via a CRDT, so setting it once on a single node is sufficient.
   ```elixir


F2011B:
I have tried to use the [Tanx example](https://github.com/ElixirSeattle/tanx) ,
to understand how a handoff can be used to persist state over changing cluster 
layout.
 I cut out the handoff part and included it into a small example project.
The :set_members message is handled by the handoff implementation to 
set the neighbours of its DeltaCRDTs.

code snippet from 
Tanx.Handoff.Impl (file handoff/impl.ex)
```elixir  
    def handle_call({:set_members, members}, _from, state) do
    Logger.info("In Set Members!!! #{inspect(members)}")

    existing_members = MapSet.new(Map.keys(DeltaCrdt.read(members_crdt_name(state.name))))
    new_members = MapSet.new(member_names(members))

    Enum.each(MapSet.difference(existing_members, new_members), fn removed_member ->
      DeltaCrdt.mutate_async(members_crdt_name(state.name), :remove, [removed_member])
    end)
```

I am not sure about the source of this message.
It seems that this message is sent
by Horde.Cluster.set_members to its members.

Maybe there is a better example for implementing state handoff.
Maybe you can clarify that with a really small example.
@derekkraan I have to note I am a total noob to elixir, therefore 
thank you for your kind response.

derekkraan:
I see where the confusion is coming from. In the Tanx example app, Daniel is taking advantage of the existing `Horde.Cluster.set_members/2` function. Buuuuut, while this function is public, I would classify the `{:set_members, members}` message as private. So I don't think it would make sense to document this here.

The Tanx example depends on delta_crdt, and I think that's really something that should be grokked before using it for state handoff like this (imho). There is a guide on state handoff, https://hexdocs.pm/horde/state_handoff.html#content, if you can suggest improvements to that then that would be much appreciated, although I would like to keep it storage-agnostic (and not go into details of how to wire up delta_crdt into a cluster, for example).

derekkraan:
Closing this issue but still happy to answer your questions here (or in the #horde channel on Slack)

PR #143, opened 2019-07-25
Ensure Registry.child_spec/1 overridable.

Ajwah:
* `child_spec` configurable with inline options during `use` like for regular `GenServer`
* `child_spec` can be redefined and refined with `super`

### DIFF
--- a/lib/horde/registry.ex
+++ b/lib/horde/registry.ex
@@ -34,34 +34,38 @@ defmodule Horde.Registry do
   Then you can use `MyRegistry.child_spec/1` and `MyRegistry.start_link/1` in the same way as you'd use `Horde.Registry.child_spec/1` and `Horde.Registry.start_link/1`.
   """
 
-  @callback init(options :: Keyword.t()) :: {:ok, options :: Keyword.t()}
-
-  defmacro __using__(_opts) do
-    quote do
-      @behaviour Horde.Registry
-
-      def child_spec(options) do
-        options = Keyword.put_new(options, :id, __MODULE__)
-
-        %{
-          id: Keyword.get(options, :id, __MODULE__),
-          start: {__MODULE__, :start_link, [options]},
-          type: :supervisor
-        }
-      end
-
-      def start_link(options) do
-        Horde.Registry.start_link(Keyword.put(options, :init_module, __MODULE__))
-      end
-    end
-  end
-
   @type option ::
           {:keys, :unique}
           | {:name, registry()}
           | {:delta_crdt_options, [DeltaCrdt.crdt_option()]}
           | {:members, [Horde.Cluster.member()]}
-          | {:listeners, [atom()]}
+
+  @callback init(options :: Keyword.t()) :: {:ok, options :: Keyword.t()}
+  @callback child_spec(options :: [option()]) :: Supervisor.child_spec()
+
+  defmacro __using__(opts) do
+    quote location: :keep, bind_quoted: [opts: opts] do
+      @behaviour Horde.Registry
+
+      def child_spec(options) do
+        options = Keyword.put_new(options, :id, __MODULE__)
+
+        default = %{
+          id: Keyword.get(options, :id, __MODULE__),
+          start: {__MODULE__, :start_link, [options]},
+          type: :supervisor
+        }
+
+        Supervisor.child_spec(default, unquote(Macro.escape(opts)))
+      end
+
+      def start_link(options) do
+        Horde.Registry.start_link(Keyword.put(options, :init_module, __MODULE__))
+      end
+
+      defoverridable child_spec: 1
+    end
+  end
 
   @type registry :: atom()
 
--- a/test/support/module_based_registry.ex
+++ b/test/support/module_based_registry.ex
@@ -13,3 +13,46 @@ defmodule TestRegistry2 do
     {:ok, Keyword.put(options, :members, [:init_test_1, :init_test_2])}
   end
 end
+
+defmodule TestRegistry3 do
+  use Horde.Registry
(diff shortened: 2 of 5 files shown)


lib/horde/registry.ex
```diff
@@ -36,23 +36,27 @@ defmodule Horde.Registry do
 
   @callback init(options :: Keyword.t()) :: {:ok, options :: Keyword.t()}
 
-  defmacro __using__(_opts) do
-    quote do
+  defmacro __using__(opts) do
+    quote location: :keep, bind_quoted: [opts: opts] do
```
Ajwah: As inspired by https://github.com/elixir-lang/elixir/blob/c76c5d6d188bbb787d6afefab74a31080695d1f2/lib/elixir/lib/gen_server.ex#L688

lib/horde/registry.ex
```diff
@@ -36,23 +36,27 @@ defmodule Horde.Registry do
...
         options = Keyword.put_new(options, :id, __MODULE__)
 
-        %{
+        default = %{
           id: Keyword.get(options, :id, __MODULE__),
           start: {__MODULE__, :start_link, [options]},
           type: :supervisor
         }
+
+        Supervisor.child_spec(default, unquote(Macro.escape(opts)))
```
Ajwah: As inspired by: https://github.com/elixir-lang/elixir/blob/c76c5d6d188bbb787d6afefab74a31080695d1f2/lib/elixir/lib/gen_server.ex#L702

Personal note: I do not understand why: `unquote(Macro.escape(opts))` as opposed to simply doing: `unquote(opts)`. The former makes it difficult to override the key: `start` MFA

lib/horde/registry.ex
```diff
@@ -36,23 +36,27 @@ defmodule Horde.Registry do
...
       end
 
       def start_link(options) do
         Horde.Registry.start_link(Keyword.put(options, :init_module, __MODULE__))
       end
+
+      defoverridable child_spec: 1
```
Ajwah: As inspired by: https://github.com/elixir-lang/elixir/blob/c76c5d6d188bbb787d6afefab74a31080695d1f2/lib/elixir/lib/gen_server.ex#L705

test/support/module_based_registry.ex
```diff
@@ -13,3 +13,45 @@ defmodule TestRegistry2 do
     {:ok, Keyword.put(options, :members, [:init_test_1, :init_test_2])}
   end
 end
+
+defmodule TestRegistry3 do
+  use Horde.Registry
+  @base_opts [keys: :unique, name: __MODULE__]
```
Ajwah: Included so as to demonstrate that valid `child_spec/1` can indeed be generated to facilitate the objective of: `Supervisor.start_link([TestRegistry3], strategy: :one_for_one)`

test/support/module_based_registry.ex
```diff
@@ -13,3 +13,45 @@ defmodule TestRegistry2 do
...
+      type: :supervisor
+    }
+  end
+end
+
+defmodule TestRegistry4 do
+  use Horde.Registry,
+    restart: :transient,
+    # value provided to differ from default :supervisor option
+    type: :worker
```
Ajwah: Demonstrate configurability of `child_spec` through use of options like [GenServer](https://github.com/elixir-lang/elixir/blob/c76c5d6d188bbb787d6afefab74a31080695d1f2/lib/elixir/lib/gen_server.ex#L128-L130)

test/uniform_distribution_test.exs
```diff
@@ -4,16 +4,20 @@ defmodule UniformDistributionTest do
 
   property "chooses one of the members" do
     member =
-      ExUnitProperties.gen all node_id <- integer(1..100_000),
-                               status <- StreamData.member_of([:alive, :dead, :shutting_down]),
-                               name <- binary(),
-                               pid <- atom(:alias) do
+      ExUnitProperties.gen all(
```
Ajwah: mix format

lib/horde/registry.ex
```diff
@@ -34,34 +34,39 @@ defmodule Horde.Registry do
...
   """
 
+  @type option ::
+          {:keys, :unique}
+          | {:name, registry()}
+          | {:delta_crdt_options, [DeltaCrdt.crdt_option()]}
+          | {:members, [Horde.Cluster.member()]}
+
   @callback init(options :: Keyword.t()) :: {:ok, options :: Keyword.t()}
+  @callback child_spec(options :: [option()]) :: Supervisor.child_spec()
```
Ajwah: Adopted as is from [code base](https://github.com/Ajwah/horde/blob/0beaaacfea51f179e7e6e855c089077e270f165a/lib/horde/registry.ex#L89). Introduced because of:
> Elixir has always advocated that a behaviour must always be defined when a set of functions is marked as overridable
as stated by [Jose Valim](https://github.com/elixir-lang/elixir/issues/5734)

derekkraan: :joy: I wrote that spec in Elixir.Registry, just got a PR for that accepted this week.

lib/horde/registry.ex
derekkraan: I'm also not sure, if it works without this, then I think that's also fine.

lib/horde/registry.ex
derekkraan: Also, could you please add this to `Horde.Supervisor`?

Ajwah:
sure I will do in separate PR

lib/horde/registry.ex
Ajwah: I will add it in a separate PR

Issue #144, opened 2019-07-25
Lookup return value when not found is not compatible with Elixr.Registry


philipbrown:
Hey @derekkraan, thank you for creating Horde. I'm just testing it out on a project and it's working great so far!

One thing I've noticed is `Horde.Registry.lookup/2` returns `:undefined` when there isn't a match, whereas `Elixir.Registry.lookup/2` returns an empty list.

I can send a PR to fix it if you like? 😄

derekkraan:
Yes please!

PR #145, opened 2019-07-25
Fix the return value of `Horde.Registry.lookup/2` to return an empty …

philipbrown:
…list when no match

Fixes https://github.com/derekkraan/horde/issues/144


derekkraan:
Looks great, thanks!!

PR #146, opened 2019-07-25
Promote typespec for child_spec/1 to callback

Ajwah:


lib/horde/supervisor.ex
```diff
@@ -58,7 +74,7 @@ defmodule Horde.Supervisor do
           type: :supervisor
         }
 
-        Supervisor.child_spec(default, unquote(Macro.escape(opts)))
+        Supervisor.child_spec(default, unquote(opts))
```
Ajwah: As inspired from: https://github.com/elixir-lang/elixir/commit/2e4d58cfa224a9604e5982758e92928f983771e2

lib/horde/supervisor.ex
```diff
@@ -69,12 +85,9 @@ defmodule Horde.Supervisor do
     end
   end
 
-  @callback init(options()) :: {:ok, options()}
-
   @doc """
   See `start_link/2` for options.
   """
-  @spec child_spec(options :: options()) :: Supervisor.child_spec()
```
Ajwah: `spec` is now redundant in presence of `callback`

derekkraan:
:ok-hand:

PR #147, opened 2019-07-25
Ensure Registry.child_spec/1 overridable

Ajwah:
minor modifications

### DIFF
--- a/lib/horde/registry.ex
+++ b/lib/horde/registry.ex
@@ -47,6 +47,7 @@ defmodule Horde.Registry do
     quote location: :keep, bind_quoted: [opts: opts] do
       @behaviour Horde.Registry
 
+      @impl true
       def child_spec(options) do
         options = Keyword.put_new(options, :id, __MODULE__)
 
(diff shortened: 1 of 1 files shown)


lib/horde/registry.ex
```diff
@@ -56,7 +57,7 @@ defmodule Horde.Registry do
           type: :supervisor
         }
 
-        Supervisor.child_spec(default, unquote(Macro.escape(opts)))
+        Supervisor.child_spec(default, unquote(opts))
```
Ajwah: As inspired from here: https://github.com/elixir-lang/elixir/commit/2e4d58cfa224a9604e5982758e92928f983771e2

lib/horde/registry.ex
```diff
@@ -86,7 +87,6 @@ defmodule Horde.Registry do
   ])
   ```
   """
-  @spec child_spec(options :: [option()]) :: Supervisor.child_spec()
```
Ajwah: redundant because of callback

derekkraan:
Thanks!!

PR #148, opened 2019-07-25
update libcluster guide

broodfusion:
@derekkraan We should also be adding our Registry to `set_members` here, correct? If so, I'll add this into the readme, so it will create less confusion for others.

```elixir
  def handle_info({:nodeup, _node, _node_type}, state) do
    set_members(MyHordeRegistry)
    set_members(MyHordeSupervisor)
    {:noreply, state}
  end

  def handle_info({:nodedown, _node, _node_type}, state) do
    set_members(MyHordeRegistry)
    set_members(MyHordeSupervisor)
    {:noreply, state}
  end
```

-- Shawn


derekkraan:
Yes. It's just an example to help you on your way, but if you think it's clearer like that then feel free to modify!

broodfusion:
Alright, I've updated the guide with registry implementation as well.

derekkraan:
:+1: thanks for contributing!

PR #153, opened 2019-08-08
handle_info :name_conflict with correct :EXIT match

markmeeus:
We encountered a small error in the guides. When trapping exits, the actual message is wrapped in an {:exit, from, reason} tuple.


derekkraan:
Thank you!!

Issue #154, opened 2019-08-14
Crash trying to shut down the ProcessSupervisor


mgwidmann:
Perhaps I'm doing something wrong so please correct me if I have.

I have 5 different registries/supervisors to split the load of my application. One for "everything else" and 4 for one particular entity which I've split up by hashing the name of the worker and starting the workers under that supervisor (and registering the name in that registry).

I'm using local-cluster with dynamic supervision (and the example nodewatching genserver) for testing where I see this the most, but it also happens when two machines connect together.

It doesn't say why the `ProcessSupervisor` shut down before this error occurred, so I assume it was a normal shutdown. Any ideas?

```
2019-08-14T05:04:31.573000 [info] [libcluster:gossip] connected to :"w2@127.0.0.1"

2019-08-14T04:51:03.580000 [error] GenServer MyApplication.DistributedSupervisor terminating
** (stop) exited in: GenServer.stop(MyApplication.DistributedSupervisor.ProcessesSupervisor, :normal, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir) lib/gen_server.ex:938: GenServer.stop/3
    (horde) lib/horde/supervisor_impl.ex:570: Horde.SupervisorImpl.shut_down_all_processes/1
    (horde) lib/horde/supervisor_impl.ex:347: Horde.SupervisorImpl.handle_info/2
    (stdlib) gen_server.erl:637: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:711: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: {:crdt_update, [{:add, {:member_node_info, {MyApplication.DistributedSupervisor, :"manager@127.0.0.1"}}, %Horde.Supervisor.Member{name: {MyApplication.DistributedSupervisor, :"manager@127.0.0.1"}, status: :shutting_down}}]}
[...]

PR #155, opened 2019-08-15
Example dynamic cluster membership with tests.

mgwidmann:
Currently the tests fail due to the issue #154 

@derekkraan Can you get the crashes to stop happening?


mgwidmann:
Have you had a chance to look at this test? I'm seeing shutdowns occur while my system is running at full load, not sure if its related but it certainly causes a lot of noise when it does since it all I can find is this same error message. Perhaps there is a further root cause I cannot find on why the system is shutting down...

derekkraan:
I haven't had a chance to look at this yet. Work is busy these days.

PR #158, opened 2019-09-05
Revert delta crdt options field naming

Shikanime:
My fault, I didn't remember when I renamed `delta_crdt_options` to `delta_crdt` in my previous pull request, and it wasn't my intention.


derekkraan:
Ok, sounds good. I was actually ok with this change, but I do have a slight preference for `delta_crdt_options`.

PR #160, opened 2019-09-17
fix spec on Cluster.members/1

pedro-gutierrez:
Minor stuff.


derekkraan:
Is this correct? If I look at SupervisorImpl and RegistryImpl, then the existing spec seems to match. Note: it's possible this is different in master vs the latest release.

pedro-gutierrez:
You are totally right -- closing this.

PR #162, opened 2019-09-19
Add the dialyxir helper tool for dialyzer

gausby:
I am working on a project that has a dialyzer step in its build pipeline, and it seems that I have various problems with code that has different specs than the actual implementation originating from Horde.

To help with ironing out these issues I have added the dialyxir helper tool to the project. For the current master it shows 3 discrepancies. To run the dialyzer task it should be as easy as running `mix dialyzer` in the root of the project.


derekkraan:
Thanks, I'll take a look at the errors and then we can also add it to the circleci build.

PR #168, opened 2019-10-11
Update getting_started.md

criticalbh:
libcluster link is not working


derekkraan:
Hi,

This link works here: https://hexdocs.pm/horde/getting_started.html#running-a-genserver

But not in the github repo I guess? I'm not sure if we can have it working in both.

criticalbh:
> Hi,
> 
> This link works here: https://hexdocs.pm/horde/getting_started.html#running-a-genserver
> 
> But not in the github repo I guess? I'm not sure if we can have it working in both.

Hi, you are right. It wont work

PR #176, opened 2019-11-08
Fix typo #2

jfrolich:


derekkraan:
Thanks, this was indeed missed in the great child_spec refactor of 2019.

Issue #171, opened 2019-10-29
Use PG2 as default cluster formation


Shikanime:
It would be more convenient to create a cluster of `DynamicSupervisor` / `Registry` using PG2 by default such as `Phoenix.Pubsub`.

derekkraan:
Can you elaborate?

Shikanime:
Actually every Horde components are discovered statically via Erlang node name:

```elixir
{HelloWorld.HelloDynamicSupervisor, :"count1@127.0.0.1"}
```

In the case of a dynamic cluster such as under a container orchestrator like Kubernetes, we have to monitor the node to join or leave and ensure proper synchronization of the Horde cluster on the user space can be sometime less convenient.

Instead the component should join the pg2 `{:horde, name}` group at startup where name can be  `HelloWorld.HelloDynamicSupervisor`.

What do you think ?

derekkraan:
I think I would rather leave this up to the person who is using Horde. If you would like to submit a PR documenting how to get Horde working with pg2, then I would accept that.

PR #184, opened 2019-11-19
Add :auto option for cluster membership

bernardd:
After realising I was about to re-implement it for the second time, I took a stab at creating a built-in version of [dynamic cluster membership support](https://hexdocs.pm/horde/libcluster.html#dynamic-cluster-membership). This allows the `members` option to be specified as `:auto` and uses the system described in the link above to keep cluster membership in sync with all visible nodes. It seemed like a common enough use case (and something I was mildly surprised to find was missing when I first picked up Horde) to warrant inclusion in the library itself.

I haven't written any docs for this yet - I figured I'd get some feedback first to see if you thought it was a worthwhile inclusion.

### DIFF
--- a/lib/horde/dynamic_supervisor_impl.ex
+++ b/lib/horde/dynamic_supervisor_impl.ex
@@ -59,7 +59,11 @@ defmodule Horde.DynamicSupervisorImpl do
   def handle_continue({:set_members, nil}, state), do: {:noreply, state}
 
   def handle_continue({:set_members, :auto}, state) do
-    Horde.NodeListener.initial_set(state.name)
+    state =
+      state.name
+      |> Horde.NodeListener.make_members()
+      |> set_members(state)
+
     {:noreply, state}
   end
 
--- a/lib/horde/dynamic_supervisor_impl.ex
+++ b/lib/horde/dynamic_supervisor_impl.ex
@@ -58,6 +58,11 @@ defmodule Horde.DynamicSupervisorImpl do
 
   def handle_continue({:set_members, nil}, state), do: {:noreply, state}
 
+  def handle_continue({:set_members, :auto}, state) do
+    Horde.NodeListener.initial_set(state.name)
+    {:noreply, state}
+  end
+
   def handle_continue({:set_members, members}, state) do
     {:noreply, set_members(members, state)}
   end
--- a/guides/libcluster.md
+++ b/guides/libcluster.md
@@ -2,7 +2,15 @@
 
 Horde doesn't provide functionality to set up your cluster, we recommend you use [`libcluster`](https://hexdocs.pm/libcluster/readme.html) for this purpose.
 
-There are two strategies you can use to integrate libcluster with Horde:
+There are three strategies you can use to integrate libcluster with Horde:
+
+## Automatic Cluster Membership
+
+When starting a `Horde.Registry` or `Horde.DynamicSupervisor`, setting the `members`
+option to have a value of `:auto` will automate membership management. In this
+mode, all visible nodes will be initially added to the cluster. In addition,
+any new nodes that become visible will be automatically added and any
+nodes that shut down will be automatically removed.
 
 ## Static Cluster Membership
 
@@ -34,7 +42,7 @@ This is the simplest approach. You tell Horde which members are supposed to be i
 
 ## Dynamic Cluster Membership
 
-If you will be adding and removing nodes from your cluster constantly, and don't want to repackage your application every time you do this, then you will need to perform a couple of extra steps.
+If you will be adding and removing nodes from your cluster constantly, and don't want to repackage your application every time you do this, then you will need to perform a couple of extra steps (assuming your needs cannot be met by the [`:auto` setting](#automatic-cluster-membership)).
 
 In this scenario, you will need to implement a [module-based Supervisor](https://hexdocs.pm/horde/Horde.DynamicSupervisor.html#module-module-based-supervisor)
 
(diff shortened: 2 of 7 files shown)


derekkraan:
Hi @bernardd , thanks for the PR. I had already sort of said I didn't want to have something like this in #171, but now I am seeing it in code it does seem reasonable. We aren't taking over the role of `libcluster`, just joining nodes into Horde whenever they connect.

So in principle, this should be fine.

The only comment I have at this point is that we would like to avoid using `Application` if at all possible. I think it should be possible to fit the `NodeListener` into the regular supervision tree for `Horde.DynamicSupervisor` and `Horde.Registry`.

derekkraan:
Hi @bernardd , thanks for the PR. I had already sort of said I didn't want to have something like this in #171, but now I am seeing it in code it does seem reasonable. We aren't taking over the role of `libcluster`, just joining nodes into Horde whenever they connect.

So in principle, this should be fine.

The only comment I have at this point is that we would like to avoid using `Application` if at all possible. I think it should be possible to get the

arjan:
If you go this way there is also some kind of whitelisting / blacklisting mechanism required, to ensure that temporary nodes do not join the horde directly (distillery's `_main_` nodes come to mind), or are those nodes invisible?

To be honest this feels a bit like implementing Swarm all over again.. ;-) I am not sure this functionality belongs in this library, it feels too high-level for me. Although it might be a "good default" when you're just starting with Horde.

For instance, in [Sworm](https://github.com/arjan/sworm/blob/master/lib/sworm/directory_manager.ex) I also do auto membership, but not just joining each node, only those where the specific Horde registries / supervisors are actually running.

bernardd:
> If you go this way there is also some kind of whitelisting / blacklisting mechanism required

No more than there was before. The existing functionality still exists and works perfectly fine, just as it did before (and still allows you to implement whatever system you so desire). This is just a simple (for some value of "sensible") default behaviour that lowers the barrier to entry just a little.

> or are those nodes invisible?

It depends entirely on the configuration of whatever created them. Since you mention distillery, though, from a glance at the code it looks like it does indeed connect as a hidden node, which is exactly what I'd expect since the control node it creates is not really "part of the cluster", and that kind of behaviour is exactly what hidden nodes are for.

> To be honest this feels a bit like implementing Swarm all over again.. ;-)

Except it isn't. It's trying to add some of the ease of use that comes with Swarm, while still allowing more sophisticated use cases if the easy default isn't appropriate. Plus, it still maintains the key distinction from Swarm (which is what made me move from it to Horde in the first place) of separating the supervision and registry functionality.

bernardd:
> The only comment I have at this point is that we would like to avoid using Application if at all possible.

Great, thanks - I'll have a look at sorting that out tomorrow.

bernardd:
> I think it should be possible to fit the NodeListener into the regular supervision tree for Horde.DynamicSupervisor and Horde.Registry.

So the problem with this is that without an application module, there's no clear "top level" location under which to hang the `NodeListener` process. I can see four options:
1. Each `Registry` or `DynamicSupervisor` (which I'll collectively call "elements") with `:auto` set starts it own `NodeListener` which it supervises. That feels like unnecessary overhead, but it would probably work okay.
2. Each element with `:auto` set checks for an existing registered `NodeListener` on the node, and starts one (which it itself supervises) if it doesn't exist. That carries the dual problems of race conditions (if two different Horde elements are started at the same time) and it not being clear where the `NodeListener` can be found in the supervision tree. There's also the question of what happens to the `NodeListener` if the element is subsequently shut down while there are still other elements relying on it.
3. Require the user explicitly start the `NodeListener` under their own app's supervision tree before bringing up any `:auto` elements. That would work, but it feels a little messy to require the user to start things that are essentially internal functionality.
4. Stick with the `Application` implementation. In the absence of any other options, I think this still gets my vote. The application module system is provided precisely to allow the app/library to start and manage internal processes without the caller having to worry about them. It means the `:auto` setting can "just work" and the `NodeListener` process is always findable in a consistent, logical location.

I'm more than happy to be talked around, though, if you have a strongly different opinion :)

derekkraan:
Hi @bernardd, I understand why you would have gone with option 4. I do think I would prefer option 1 however. The additional overhead will be minimal, and it'll be a bit simpler.

derekkraan:
I do think this PR is something that will help people get up and running more quickly with Horde, so I will merge it when you make the change to option 1. I think it'll cover 95% of people's use cases more or less.

I do understand the comparisons to Swarm, @arjan, but I think in this case we're still doing the right thing. Since Swarm was implemented as an Application, you could only get one instance per node (or one swarm per cluster). This PR doesn't change that you can have multiple hordes per node / cluster.

bernardd:
Yep, no worries. You could also make the argument that it may scale better since node change management could be done concurrently for all the `:auto` elements. I'll make that change on Monday.

bernardd:
There we go - let me know if there's anything else you'd like tweaked.

lib/horde/dynamic_supervisor_impl.ex
```diff
@@ -58,6 +58,11 @@ defmodule Horde.DynamicSupervisorImpl do
 
   def handle_continue({:set_members, nil}, state), do: {:noreply, state}
 
+  def handle_continue({:set_members, :auto}, state) do
+    Horde.NodeListener.initial_set(state.name)
```
derekkraan: I'm unsure about this bit right here. This will result in a race condition. It's possible that some requests will come in to this supervisor before it's had a chance to set its members.

bernardd: Ah, good catch. I've fixed it to it a synchronous part of the supervisor startup (and also refactored the `Horde.Registry` initialisation to do the same, if only for consistency).

derekkraan:
:muscle: thanks for the contribution @bernardd !

sikanrong:
Hey all just wanted to chime in and say, I have almost exactly this implemented in my own code, and it does seem reasonable to me that it would be included as a part of the library. I'm excited to start using this!

PR #185, opened 2019-11-21
Add conflict_fun option

bernardd:
So I was doing some work with the `Horde.Registry` stuff and hit a bit of an issue: If the registered process is a supervisor, then firing the `{:name_conflict, ...}` exit at it doesn't actually cause it to exit - instead it's silently discarded. And worse, if you're using an OTP supervisor, you can't intercept the message to cause the exit yourself. So after the conflict you're left with two supervisors still running, but only one registered.

I played around with a bunch of different ideas, but in the end the neatest one seemed to be to give the option of dealing with name conflicts yourself. So this change adds a `conflict_fun` option to `Horde.Registry`'s startup which replaces the exit with a call to a function (providing the same information that the exit contains). The user can then do whatever is required to shut down the conflict-losing process (and anything else they might want to do).

I did look at the undocumented (as far as I could see) `listeners` option - listening for the `:unregister` message would work, but it is also impossible to distinguish between an `:unregister` due to a name conflict and one due to any other cause.

If you're happy with this I'll add some documentation. If you'd prefer another approach, I'd be quite happy to code up anything that makes sense and solves the problem.

### DIFF
--- a/lib/horde/registry.ex
+++ b/lib/horde/registry.ex
@@ -44,6 +44,7 @@ defmodule Horde.Registry do
   @type option ::
           {:keys, :unique}
           | {:name, atom()}
+          | {:conflict_fun, (pid(), pid(), {atom(), term()}, atom() -> any())}
           | {:delta_crdt_options, [DeltaCrdt.crdt_option()]}
           | {:members, [Horde.Cluster.member()]}
 
@@ -98,6 +99,7 @@ defmodule Horde.Registry do
       :listeners,
       :meta,
       :keys,
+      :conflict_fun,
       :distribution_strategy,
       :members,
       :delta_crdt_options
@@ -127,6 +129,7 @@ defmodule Horde.Registry do
     end
 
     listeners = Keyword.get(options, :listeners, [])
+    conflict_fun = Keyword.get(options, :conflict_fun, nil)
     meta = Keyword.get(options, :meta, nil)
     members = Keyword.get(options, :members, [])
     delta_crdt_options = Keyword.get(options, :delta_crdt_options, [])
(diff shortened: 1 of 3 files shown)


derekkraan:
Hi @bernardd,

I am wondering what the use case of registering a supervisor would be. As you say, Elixir.Supervisor is very limited in terms of allowing the programmer to add behaviours. 

The PR itself, code-wise, looks fine. I am just not sure about the actual utility of this change.

Derek

bernardd:
So by way of explanation - I'm writing a library to make it easy to create a singleton process (or, importantly, tree of processes) - by which I mean a process that exists exactly once across all nodes. The mechanism for ensuring there is (eventually) only one instance is, of course, `Horde.Registry`. It all works very nicely when the singleton is just a simple GenServer, but if I want a bunch of processes with a particular restart type (`:rest_for_one`, for example) then obviously they need to go under a supervisor. And the only way that really makes sense to do that (that I've come up with) is to have that top-level supervisor as the registered process. But if it doesn't exit when it receives the `:name_conflict` message, then I end up with multiple of them running, but only one registered with the `Horde.Registry`.

derekkraan:
So one possible way to work around this would be to have another process in your supervision tree that you then register with `Horde.Registry`, which is then responsible for shutting down the supervision tree if it receives a `:name_conflict` exit message.

It is in any case tricky to register a supervisor process. The real reason the Registry sets up a link to a registered process is so that if the registry itself dies, then the process will be notified about this and be able to take appropriate action. The usual course of action is to just allow the process to die and then upon restarting, register it again with the (now restarted) registry.

By registering a supervisor, however, you're breaking this behaviour, and you'll run the risk of having supervisor processes that are not registered in the registry.

So this PR won't be a real fix for your problem. The link (aka exit) behaviour needs to remain intact.

bernardd:
Ah, I hadn't thought about the registry dying - that's a really good point. Okay, I'll go away and play with this some more and come up with a better idea - thanks heaps for the feedback :)

PR #187, opened 2019-12-04
Adds Horde.Registry.delete_meta/2

weakwire:


derekkraan:
Hi @weakwire , thanks for the PR. Can you tell me why you need this feature? (And why, for example, `put_meta(key, nil)` is not sufficient).

weakwire:
You are right. I was not aware that put_meta(key, nil) cleans up the key from the registry. I assumed it just assigns a new value. Maybe a documentation update here would be useful.

derekkraan:
@weakwire Could you submit a PR to improve the documentation as you've described?

Thanks
Derek

Issue #181, opened 2019-11-14
DynamicSupervisor and Registry init spec is wrong.


deadtrickster:
Example: 

```elixir
<...>/horde/supervisor.ex:10:callback_type_mismatch
Type mismatch for @callback init/1 in Horde.DynamicSupervisor behaviour.

Expected type:

  {:ok,
   [
     {:delta_crdt_options
      | :distribution_strategy
      | :extra_arguments
      | :max_restarts
      | :max_seconds
      | :members
      | :name
      | :shutdown
      | :strategy, atom() | [any()] | integer()}
   ]}


Actual type:

  :ignore
  | {:ok,
     {%{:intensity => _, :period => _, :strategy => _}, [any()]}
     | %{
         :delta_crdt_options => %{:max_sync_size => _, :shutdown => _, :sync_interval => _},
         :distribution_strategy => _,
         :extra_arguments => _,
         :intensity => _,
         :max_children => _,
         :members => _,
         :period => _,
         :strategy => _
       }}
  | {:stop, {:bad_return, {atom(), :init, _}}}

```

```elixir
  def init(options) do
    [strategy: :one_for_one, members: get_members()]
    |> Keyword.merge(options)
    |> Horde.DynamicSupervisor.init()
  end
```

anthonator:
Has this been resolved by #189 or are there still typespec issues?

derekkraan:
I think this one has been resolved @anthonator

gbrlcustodio:
Actually, #189 fixes `DynamicSupervisor`'s wrong specs. It still complains about `Registry` one.

Expected:

```elixir
{:ok, Keyword.t()}
```

Got:

```elixir
:ignore
| {:ok,
    {%{intensity: any, period: any, strategy: any}, [any]}
    | %{
        delta_crdt_options: %{
          max_sync_size: any,
          shutdown: any,
          sync_interval: any
        },
        distribution_strategy: any,
        keys: any,
        listeners: any,
        members: any,
        meta: any
      }}
| {:stop, {:bad_return, {atom, :init, any}}}
```

anthonator:
Anybody know how to suppress this warning until 381279c has been released?

gbrlcustodio:
Hey @anthonator 

I'm using `@dialyzer {:nowarn_function, init: 1}` to disable it inside my module.

anthonator:
Ah, nice! I was making this more complicated than it needed to be. Thank you!

PR #189, opened 2019-12-17
Fix Dynamic Supervisor Init typespec - Issue #181

scudelletti:
Hello! 👋 

This PR fixes issue #181.

Currently horde has a few typespecs errors and one of them is in the init function inside the `Horde.DynamicSupervisor`.

This fix reduces the total amount of typespec errors from 10 to 6.
To see the errors I suggest you to run dialyzer in the test environment(`MIX_ENV=test mix dialyzer`) since the tests contain code that uses the library giving more context to dialyzer. 

Thanks for you time and thank you very much for maintaining Horde! It's an awesome library! 🎉

### DIFF
--- a/lib/horde/dynamic_supervisor.ex
+++ b/lib/horde/dynamic_supervisor.ex
@@ -65,7 +65,11 @@ defmodule Horde.DynamicSupervisor do
           | {:delta_crdt_options, [DeltaCrdt.crdt_option()]}
           | {:process_redistribution, :active | :passive}
 
-  @callback init(options()) :: {:ok, options()}
+  @callback init(options()) ::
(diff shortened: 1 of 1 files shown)


derekkraan:
Thanks for the PR. I have simplified the spec to `{:ok, options()} | :ignore`. I'm not sure what dialyzer meant by adding all those other extra things to the callback, but this also seems to satisfy dialyzer.

anthonator:
Would it be possible to get a patch release of this?

KamilZielinski:
@anthonator No patch so far but there's `0.8.0-rc.1` which seems to contain these changes. You may want to try it out

PR #194, opened 2020-02-21
Gracefully handle exit signals when stopping.

### LINKED ISSUE #193: Unhandled error on SIGTERM
I noticed this in my logs on shutdown:

```
[info] SIGTERM received - shutting down

[error] GenServer Service.Scheduler.Supervisor terminating
** (stop) exited in: GenServer.stop(Service.Scheduler.Supervisor.ProcessesSupervisor, :normal, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir 1.10.1) lib/gen_server.ex:981: GenServer.stop/3
    (horde 0.7.1) lib/horde/dynamic_supervisor_impl.ex:588: Horde.DynamicSupervisorImpl.shut_down_all_processes/1
    (horde 0.7.1) lib/horde/dynamic_supervisor_impl.ex:365: Horde.DynamicSupervisorImpl.handle_info/2
    (stdlib 3.11.2) gen_server.erl:637: :gen_server.try_dispatch/4
    (stdlib 3.11.2) gen_server.erl:711: :gen_server.handle_msg/6
    (stdlib 3.11.2) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
```

It seems to be harmless, however it does make a mess of the logs.

I think [the function here](https://github.com/derekkraan/horde/blob/master/lib/horde/processes_supervisor.ex#L526) should gracefully handle the `GenServer.stop/3` error with a try/catch in the event the supervisor has already stopped.

Something like this could work:

```elixir
  def stop(supervisor, reason \\ :normal, timeout \\ :infinity) do
    try do
      GenServer.stop(supervisor, reason, timeout)
    catch
      # process doesn't exist, ignore error
      :exit, {:noproc, _err} -> :ok
    end
  end
```

brndnmtthws:
This should resolve #193.


brndnmtthws:
Closing for now, because there's some more work to do. I'll open a new PR when I'm ready.

derekkraan:
@brndnmtthws did you mean to close this PR?

PR #199, opened 2020-03-15
Replace remaining references to Horder.Supervisor

anthonator:
Replaces remaining documentation references of `Horde.Supervisor` with `Horde.DynamicSupervisor`. Found this to be confusing when attempting to use `horde` for the first name.


derekkraan:
Thank you!!

PR #201, opened 2020-04-27
Fix Registry Init typespec - Issue #181

gbrlcustodio:
Updates `Horde.Registry.init/1` typespec as described here https://github.com/derekkraan/horde/issues/181#issuecomment-620112186

Relates to #181


derekkraan:
:+1: thanks for the fix!

Issue #203, opened 2020-04-29
Horde.Registry doesn't remove keys from removed members


balena:
Consider the following supervision tree:
```
    children = [
      {Horde.Registry, keys: :unique, name: ClusterRegistry, members: :auto},
      {SomeModule, name: {:via, Horde.Registry, {ClusterRegistry, local_key}}},
      ...
    ]
```
Your `SomeModule` is just a simple `GenServer` module. Consider the following one:
```
defmodule SomeModule do
  use GenServer

  def start_link(args),
    do: GenServer.start_link(__MODULE__, nil, name: Keyword.fetch!(args, :name))

  @impl true
  def init(_), do: {:ok, []}
end
```
Say that you're using `Cluster.Strategy.Gossip` strategy from `libcluster`, but the case applies even if you don't use `libcluster`.

Start two instances side by side and connect them. Consider the `local_key` unique for each instance. While both nodes are connected, you shall see:
```
iex> Horde.Registry.select(ClusterRegistry, [{{:"$1", :"$2", :"$3"}, [], [{{:"$1", :"$2", :"$3"}}]}])
[
  {{SomeModule, "key1"}, #PID<0.480.0>, :"host1@127.0.0.1"},
  {{SomeModule, "key2"}, #PID<35396.467.0>, :"host2@127.0.0.1"}
]
```
Which is expected. Now, drop one of the two instances and from the remaining one repeat the command above:
```
iex> Horde.Registry.select(ClusterRegistry, [{{:"$1", :"$2", :"$3"}, [], [{{:"$1", :"$2", :"$3"}}]}])
[
  {{SomeModule, "key1"}, #PID<0.480.0>, :"host1@127.0.0.1"},
  {{SomeModule, "key2"}, #PID<35396.467.0>, :"host2@127.0.0.1"}
]
```
Which is not expected.

~It's interesting to note that if you pass a `listeners` parameter to `Horde.Registry`, it _does send_ a `{:unregister, registry, key, pid}` to the listener, however the key isn't removed from the registry.~ (That's not correct, a second analysis showed just `{:register, registry, key, pid, value}` coming.

In order to circumvent this problem I created a `Janitor` process to garbage collect these keys from `Horde.Registry` when a node exits the cluster. But I ask: is this behaviour "by design"? Wouldn't it `Process.monitor/1` registered processes to avoid leaving garbage in the registry forever?

balena:
[...]

PR #204, opened 2020-04-29
Fix cleaning keys after a Horde.Registry member exits the cluster

balena:
The shortest solution is to augment the `pids_ets_table` with the member name, so on `:remove` the keys are also cleaned up.

Issue is documented here: https://github.com/derekkraan/horde/issues/203

### DIFF
--- a/lib/horde/registry.ex
+++ b/lib/horde/registry.ex
@@ -313,9 +313,9 @@ defmodule Horde.Registry do
   @spec keys(registry :: Registry.registry(), pid()) :: [Registry.key()]
   @doc "See `Registry.keys/2`."
   def keys(registry, pid) when is_atom(registry) do
-    case :ets.lookup(pids_ets_table(registry), pid) do
+    case :ets.match(pids_ets_table(registry), {{:_, pid}, :"$1"}) do
       [] -> []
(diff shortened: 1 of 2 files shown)


derekkraan:
Hi @balena, do you think you could add a test for this new code? Sorry it has taken me so long to respond, this PR looks great, I'm not sure why this didn't occur to me before :thinking:

balena:
OK, I'll do that in my spare time. However, I'm wondering: as a matter of consistency, `Horde.Supervisor` would have something defined for this case as well... Processes should be forcefully handed off / terminated. What do you think?

derekkraan:
@balena I believe there are also dangling references left in `Horde.DynamicSupervisor` when a node is removed from the members list, so we should indeed probably also clean that up.

`Horde.DynamicSupervisor` actually never force-stops a process (unless asked to by the user, or rebalancing or shutting down). Conflict detection is left up to `Horde.Registry`, because otherwise you get race conditions from enforcing uniqueness in two places.

derekkraan:
Hi @balena I've incorporated the crux of these changes in https://github.com/derekkraan/horde/commit/0779b2b07cbcce8ad58517641ab622de915ed036

PR #208, opened 2020-09-03
Add `Horde.Registry.delete_meta/2`

vhf:
[`Registry.delete_meta/2`](https://github.com/elixir-lang/elixir/pull/10298) is coming to Elixir 1.11.0 but I don't think we need to wait for 1.11.0 to be released for this feature to already land in Horde :)

### DIFF
--- a/lib/horde/registry_impl.ex
+++ b/lib/horde/registry_impl.ex
@@ -181,7 +181,7 @@ defmodule Horde.RegistryImpl do
   end
 
   defp process_diff(state, {:remove, {:registry, key}}) do
-    :ets.match_delete(state.name, {key, :_})
+    :ets.delete(state.name, key)
 
     state
   end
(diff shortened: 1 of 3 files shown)


lib/horde/registry_impl.ex
```diff
@@ -327,6 +333,18 @@ defmodule Horde.RegistryImpl do
     {:reply, :ok, state}
   end
 
+  def handle_call({:delete_meta, key, pid}, _from, state) do
+    DeltaCrdt.mutate(crdt_name(state.name), :remove, [{:registry, key}], :infinity)
+
+    for listener <- state.listeners do
+      send(listener, {:delete_meta, state.name, key, pid})
```
derekkraan, line 340: This isn't something that Elixir.Registry does so we should also not do it in Horde.

lib/horde/registry_impl.ex
```diff
@@ -180,6 +180,12 @@ defmodule Horde.RegistryImpl do
     %{state | members: new_members, nodes: new_nodes}
   end
 
+  defp process_diff(state, {:remove, {:registry, key}}) do
+    :ets.match_delete(state.name, {key, :_})
```
derekkraan, line 184: will `:ets.delete(state.name, key)` work here? I think it should

derekkraan:
Thanks for this! I'll push out 0.8.1 immediately with these changes.

Issue #215, opened 2020-10-20
Specify options to the regular Supervisor wrapping DynamicSupervisor processes


stuartish:
### The Problem

The process supervising the DynamicSupervisorImp itself is not currently getting accurate info about how many times DynamicSupervisorImpl (or another part of the system) has restarted. The cryptic layer of supervision is introduced on line 217 of dynamic_supervisor.ex. Essentially, if a child process reaches max_restarts within max_seconds, this causes the DynamicSupervisorImpl to crash. But this can happen 3 times within 5 seconds before the (regular) supervisor wrapping DynamicSupervisorImpl will actually crash, and thus be visible to its own supervising process, and there is currently no way to change that.

### Would Be Nice

If we could include the horde supervisor within a supervision tree that uses :one_for_rest, where the dynamic supervisor is the first child, to add, for instance, some default children.

### Possible Solution

One solution would be to include `max_restarts: 0` on line 217 when initializing the supervisor. This would collapse the supervision tree and always propagate crashes of DynamicSupervisorImpl, more closely emulating DynamicSupervisor behavior.

There's a less "breaking changes" type of solution tho...my proposal:
```
From 9ec161b6478b17627971cff1bdcf997da09ff6b0 Mon Sep 17 00:00:00 2001
From: Stuart Ballard <stuartrballard@gmail.com>
Date: Mon, 19 Oct 2020 20:17:12 -0500
Subject: [PATCH] to allow setting max_restart to 0

---
 lib/horde/dynamic_supervisor.ex | 10 ++++++++--
 1 file changed, 8 insertions(+), 2 deletions(-)

diff --git a/lib/horde/dynamic_supervisor.ex b/lib/horde/dynamic_supervisor.ex
index 3d79e30..8de84f9 100644
--- a/lib/horde/dynamic_supervisor.ex
+++ b/lib/horde/dynamic_supervisor.ex
@@ -149,6 +149,11 @@ defmodule Horde.DynamicSupervisor do
     delta_crdt_options = Keyword.get(options, :delta_crdt_options, [])
     process_redistribution = Keyword.get(options, :process_redistribution, :passive)

+    supervisor_options =
+      options
+      |> Keyword.get(:supervisor_options, [])
+      |> Keyword.merge(strategy: :one_for_all)
+
     distribution_strategy =
       Keyword.get(
         options,
@@ -165,7 +170,8 @@ defmodule Horde.DynamicSupervisor do
       distribution_strategy: distribution_strategy,
[...]

PR #216, opened 2020-10-21
regular supervisor wrapping dynamic supervisor always propagate restarts

stuartish:
connect #215

cc: @derekkraan


derekkraan:
Hmm, I think we're only halfway there with this PR. What we want is to be able to specify in `child_spec/1` of `Horde.DynamicSupervisor` the max_restarts and max_seconds, such that when this limit is reached for children added with `Horde.DynamicSupervisor.add_child/3`, the `Horde.DynamicSupervisor` dies.

With this change, the max_restarts and max_seconds is always 5/3. Thoughts?

stuartish:
@derekkraan I added max_restarts and max_seconds to the options for the ProcessesSupervisor.

Also:
- changed some names so we weren't going from max_seconds to intensity and then back to max_seconds 🤷
- removed duplicate `strategy` from options passed to DynamicSupervisorImpl
- added tests

PR #222, opened 2021-01-26
Implement ETS's as DynamicSupervisor internal data structures instead of Maps

dbuos:
Hi, I've been using horde, in a project where I need a huge number of processes been supervised and distributed across nodes, when this number increase I note some performance degradation.. One of my guess when I inspected the DynamicSupervisorImpl were that maybe changing the maps processes_by_id and process_pid_to_id for an ETS, could have a positive effect in the performance..

Of course after the modifications, and do some tests, this change shows a improvement in the performance of spawning new childs in the supervisor, some numbers:

image

image

I tried to keep this change as local as posible thus not modifying anything outside this internal implementation detail.

I Hope this could help to enhance a bit this fantastic library.

### DIFF
--- a/lib/horde/dynamic_supervisor_impl.ex
+++ b/lib/horde/dynamic_supervisor_impl.ex
@@ -16,8 +16,8 @@ defmodule Horde.DynamicSupervisorImpl do
   defstruct name: nil,
             members: %{},
             members_info: %{},
-            processes_by_id: %{},
-            process_pid_to_id: %{},
+            processes_by_id: nil,
+            process_pid_to_id: nil,
             local_process_count: 0,
             waiting_for_quorum: [],
             supervisor_ref_to_name: %{},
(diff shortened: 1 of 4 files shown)


derekkraan:
Hi @dbuos can you help me understand why this is faster than just using bare maps?

dbuos:
> Hi @dbuos thanks for the PR. Can you help me understand why this is faster than just using bare maps?

Hi @derekkraan, I don't have all the specific implementation details of Erlang ETS's but I've observed that:

-  For small data sets bare maps are slightly more efficient (the memory doesn't have to be copied to the process memory)
-  For large and very large data sets it seems ETS's could be more efficient because of no garbage collector overhead and the possibility of doing lookups in constant time O(1) (depends on the table type).
- Bare maps are implemented as an ordered collection thus having a lookup cost of something like O(log N).

derekkraan:
Merged, this looks great, thanks for the PR @dbuos

Issue #207, opened 2020-06-28
Removing nodes creates multiple processes out of 1 process


KamilZielinski:
Hey guys. I was learning how the library works and encountered weird issue which I can't really understand nor find explanation in the docs: **why having 1 process created within 3-nodes cluster can scale up to 8 processes after removing 2 nodes from the cluster**?

### Setup
```
Erlang/OTP 22 [erts-10.7] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]
IEx 1.10.2 (compiled with Erlang/OTP 22)
```

Deps (I'm using 0.8.0-rc.1 as it solved the issue with [init/1):](https://github.com/derekkraan/horde/pull/189)
```
{:phoenix, "~> 1.5.3"},
{:phoenix_ecto, "~> 4.1"},
{:ecto_sql, "~> 3.4"},
{:postgrex, ">= 0.0.0"},
{:phoenix_live_dashboard, "~> 0.2.0"},
{:telemetry_metrics, "~> 0.4"},
{:telemetry_poller, "~> 0.4"},
{:gettext, "~> 0.11"},
{:jason, "~> 1.0"},
{:plug_cowboy, "~> 2.0"},
{:horde, "~> 0.8.0-rc.1"}
```

### Issue
I have a DynamicSupervisor(_members: :auto_) which creates 1 process (transient, called _Conversation_ in my example) on a API request with random name. I don't worry about moving the state between nodes for now. Process always lasts 60s just waiting and then dies. 
My setup: 3 nodes (Phoenix APIs) started locally on different ports (_Node1,2,3 from left to right on the screenshot_).

1. Node1 takes the request (as I hit API's url with its port) and is supposed to create 1 process. It's being created on Node3. All available nodes show number of processes: 1
2. Few seconds later I'm sending SIGTERM to Node3 from different console. Node2 schedules 2 processes instead of 1. All available nodes show number of processes: 2
3. Few seconds later i kill the app with ctrl+c in Node2 console. Node1 schedules 8 processes instead of 1 or 2 (or even 4 as I'd assume based on previous example). Node1 is showing 8 processes running.

<img width="1440" alt="Zrzut ekranu 2020-06-28 o 03 37 20-kopia" src="https://user-images.githubusercontent.com/6136794/85936470-7ce66f00-b8fb-11ea-85d3-c5e238a65709.png">

philipgiuliani:
[...]

PR #226, opened 2021-04-07
Stop randomize_child_id when handoff

kunna:
resolves https://github.com/derekkraan/horde/issues/207

### Debug
Following the instruction from https://github.com/derekkraan/horde/issues/207, I noticed that new processes are keep added to the `DeltaCrdt.read(Conv.ConversationSupervisor.Crdt)`. Whenever a new process is handed off to another node, a duplicate record is created, with a new child.id

### Change
By removing the `randomize_child_id` during the handoff process, I was able to fix this issue.

### Question
I try to understand why we needed to add `randomize_child_id` and found this PR. https://github.com/derekkraan/horde/pull/196
This PR says, "Deduplication will be the sole responsibility of Horde.Registry", but I was not able to find the logic from the `Horde.Registry` fixing the duplicated processes. 

I think PR #196 was meant to remove `Horde.ProcessesSupervisor.terminate_child_by_id`. https://github.com/derekkraan/horde/pull/196/files#diff-36c0ba7f42365d4eeaa857f4b899986cb28d29ff94de730fd8c1934c872a0389L476

because this change stoped terminating the old process, we need a consistent id for `Horde.Registry` to prevent creating a duplicated process.

### DIFF
--- a/lib/horde/dynamic_supervisor_impl.ex
+++ b/lib/horde/dynamic_supervisor_impl.ex
@@ -423,7 +423,7 @@ defmodule Horde.DynamicSupervisorImpl do
 
               case current_member do
                 %{status: :dead} ->
-                  {_response, state} = add_child(randomize_child_id(child_spec), state)
+                  {_response, state} = add_child(child_spec, state)
 
                   state
 


bernardo-martinez:
this makes sense to solve #207 can i know why was this closed?

arjan:
I agree.

without this change it is effectively not possible to run a Horde.DynamicSupervisor *without* a Horde.Registry...

derekkraan:
I am not sure if it is possible to run a Horde.DynamicSupervisor without a Horde.Registry even with this change.

arjan:
Not sure if I follow.. why would that be? Assuming that this change fixes the process duplications on membership changes..

derekkraan:
@arjan because there are other situations in which you can end up with duplicate processes. A netsplit, for example.

derekkraan:
In a previous version, Horde.DynamicSupervisor was deduping processes, but this resulted in race conditions with Horde.Registry, resulting in _both_ processes being removed sometimes. That's why Horde.DynamicSupervisor doesn't dedupe processes anymore.

PR #238, opened 2021-11-17
Pass along max_children config to ProcessesSupervisor

dbernazal:
The `DynamicSupervisor` was not respecting the `max_children` config when starting children as the value was not being propagated to the `ProcessesSupervisor`. This PR passes along the value so the `ProcessesSupervisor` can enforce the setting.


derekkraan:
Thanks and apologies for the delay in merging!

Issue #230, opened 2021-05-24
Horde Process topology


sleipnir:
Hello recently I had the need to start a process on a specific Node and still make this process managed by Horde.DynamicSupervisor and Horde.Registry, unfortunately it is not currently possible to register a process on a specific Node via Horde. It might be possible to accomplish this by creating a customized Horde.DistributionStrategy strategy, but the signature of choose_node did not seem flexible to me so that I could provide the correct topology information expected to start the process at a specific Node.
Would it be possible to accomplish this in any way?

Thanks!

arjan:
`choose_node` receives all members of the cluster plus the identifier, so that means that you could just pick one of the members based on the identifer?

sleipnir:
@arjan Yes I agree. As long as I know at that point which identifier to choose. We would have to receive some information about which preferred Node the user wants to connect this Process

sleipnir:
This could be passed as an option to the DynamicSupervisor of the process.

redrabbit:
@sleipnir, I'm in the same boat. Currently looking for a way to start my processes on a specific node. My app require a specific node based on volume id (disk storage). Did you find a way to accomplish your goal?

@arjan, it seems that the `identifier` given to `choose_node/2` is the hash of the `child_spec` (without the `:id`) in `Horde.DynamicSupervisor.start_child/2`. In my case it's not feasible because the params given to `:start` would differ between processes.

https://github.com/derekkraan/horde/blob/c00f77d080152cc160af390069169733f1afd0d8/lib/horde/dynamic_supervisor_impl.ex#L754

I think the changes made in #128 and being a fix for compatibility issues with `Elixir.DynamicSupervisor` broke the way of defining custom distribution strategies based on a specific identifier (or at least making it very cumbersome).

sleipnir:
> @sleipnir, I'm in the same boat. Currently looking for a way to start my processes on a specific node. My app require a specific node based on volume id (disk storage). Did you find a way to accomplish your goal?
> 
> @arjan, it seems that the `identifier` given to `choose_node/2` is the hash of the `child_spec` (without the `:id`) in `Horde.DynamicSupervisor.start_child/2`. In my case it's not feasible because the params given to `:start` would differ between processes.
> 
> https://github.com/derekkraan/horde/blob/c00f77d080152cc160af390069169733f1afd0d8/lib/horde/dynamic_supervisor_impl.ex#L754

[...]

PR #239, opened 2021-11-24
Replaced "identifier" parameter from `Horde.DistributionStrategy.choose_node` for "child_spec"

CimimUxMaio:
As discussed in #230 , having `child_spec` as a parameter would be much more useful for creating custom distribution strategies.

The `identifier` parameter seems to be only related to `Horde.UniformDistribution` strategy's logic, and I think its creation would be better placed inside that module rather than in:

https://github.com/derekkraan/horde/blob/63fd3c0e679cfe4e10de1a174083e991d0b38fd1/lib/horde/dynamic_supervisor_impl.ex#L753-L760


sleipnir:
ping @derekkraan 
This is a very good feature to be merged

derekkraan:
Thanks for drawing my attention to this PR again @sleipnir and thanks for submitting @CimimUxMaio even if it took me 2 years to merge.

derekkraan:
This change is going to require moving from 0.8 to 0.9 (major minor version) when it is released, since backwards compatibility will be broken.

PR #242, opened 2022-01-07
Adds node listeners to the end of child list in :auto mode

autodidaddict:
Per discussions with @derekkraan - this appends rather than prepends the node listener so that the children can start _before_ it.

The two extraneous files in this diff are from running `mix format`


autodidaddict:
Pinging @derekkraan  - is this able to be merged?

PR #245, opened 2022-02-10
Add optional node metadata

sswrk:
I think it would be cool to be able to add some extra data to DynamicSupervisor.Member struct as it might be helpful to have some additional info besides name and state. For example, when creating a custom distribution strategy, a member could be chosen by its metadata (for example its region).


sleipnir:
@sswrk This PR was very interesting

PR #246, opened 2022-02-23
Fix crashes related to missing child ids in lookup table

arjan:
I noticed that when a cluster is restarting, Horde tends to crash
because it does some administration that is incorrect. The child
lookup for the linked ETS table can fail but this was not handled on
all places in the Dynamic Supervisor implementation.

This fix addresses three places that resulted in crashes because `nil`
was not handled from the get_item / pop_item calls.

Note that there were several other places where this was already
addressed.

### DIFF
--- a/lib/horde/dynamic_supervisor_impl.ex
+++ b/lib/horde/dynamic_supervisor_impl.ex
@@ -431,6 +431,7 @@ defmodule Horde.DynamicSupervisorImpl do
 
               case current_member do
                 %{status: :dead} ->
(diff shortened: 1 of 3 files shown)


derekkraan:
Do we know why this happens? Can you write it out in steps?

I want to understand the scenario that is leading to these references being missing before I merge this.

Sorry for taking so long to respond, I moved house a week and a half ago so life has been busy.

derekkraan:
Is this PR superceded by your new one?

arjan:
No this is a different one.

arjan:
Hmm I see I messed up, I should pushed a commit here from the other PR..

Issue #251, opened 2022-05-13
Default random node distribution strategy


l3nz:
I think it would be interesting to have a random distribution strategy "in the box". I would select the n least-used nodes and place the new node randomly on one of them - e.g. place randomly but on the least-used part of the cluster.

If you think this could be interesting/useful, I can send you a PR.

derekkraan:
Hi @l3nz In principle I am in favour of including such a distribution strategy out of the box.

l3nz:
I wrote a quick Random strategy (that may be useful, so that processes are not "pinned" to a node that happens to be un-healthy for them, e.g. because of an external rate-limit), but I think that it would be nice to change DistributionStrategy so that:

- it receives a `child_spec` instead of an `id` - it may compute the id on its own if it's needed - kind of what happens in https://github.com/derekkraan/horde/issues/230
- the list of members includes information on current processes running there. This would allow for smarter load-balancing.  *Not sure how to get that in a way that plays along with the rest of Horde, and I need some guidance there.*

If we had both things, we could implement e.g "anti-affinity" strategies - that is, place my process on a node that is in a different node or rack from that other process, so that they don't fail at the same time. Or place resource pool limits - no more that 4  processes running (say) ffmpeg per node. 

On the other hand, changing that interface means that all custom-made DistributionStrategy will break (but we can make a short guide).

derekkraan:
I am wondering whether it makes sense to change Horde to enable a distribution strategy. The reason I am wondering this is that it seems that there are many possible distribution strategies that one could want. "n least-used nodes" is one. "n least-used nodes (CPU)" is another. Will we add more sophisticated tracking to Horde every time someone needs a new distribution strategy?

I will also note that it is possible to write a distribution strategy like "n least-used nodes" without changing the internals of Horde.

l3nz:
I agree - but I think that giving a hash as the id of the process to start is a mistake, because a distribution strategy may need additional data to make an intelligent decision.  If I was to make a patch, would you accept it?

derekkraan:
The issue I have is that we are talking about changing the signature not to make something _possible_, but to make it _easy_, and this is for code that has not been written yet. I think we should write those distribution strategies first, and then later decide whether changing the signature of distribution_strategy is useful or not.

PR #252, opened 2022-05-14
Added new UniformRandomDistribution strategy

l3nz:
I added a very simple Random distribution strategy, and its test.

Most of the changes were in the ExDoc part, so that available strategies are linked from within Horde.DistributionStrategy, in the perspective of adding more in the future.

Fixes https://github.com/derekkraan/horde/issues/251 though it does not change the interface of DistributionStrategy so it's a bit more limited - but this requires a separate topic.


derekkraan:
Thanks for the PR!

PR #270, opened 2024-06-14
added child_restart_matches_dynamic_supervisor.exs

gmdorf:
Found a bug where Horde.DynamicSupervisor shuts down when single worker terminates.


derekkraan:
See my comment on the linked issue.

Issue #269, opened 2024-06-14
[BUG]: Horde Supervisor behaves one_for_all no matter what


gmdorf:
According to Supervisor docs, strategy: :one_for_one with automatic_shutdown: :never (the default, made explicit for the sake of illustration) should mean that if one child process stops, the rest of the children and the supervisor should continue running.

Assume that we have this minimal GenServer module defined:
```
defmodule BasicGenServer do
  use GenServer

  def start_link(arg) do
    GenServer.start_link(__MODULE__, arg)
  end

  @impl true
  def init(init_arg) do
    {:ok, init_arg}
  end

  @impl true
  def handle_info(:stop_normal, state) do
    {:stop, :normal, state}
  end
end
```
Running this basic test with `DynamicSupervisor` succeeds as expected:
```
{:ok, sup} = DynamicSupervisor.start_link([strategy: :one_for_one, name: TestSup, automatic_shutdown: :never])

{:ok, child1} = DynamicSupervisor.start_child(sup, %{id: :child1, start: {BasicGenServer, :start_link, [0]}})
{:ok, child2} = DynamicSupervisor.start_child(sup, %{id: :child2, start: {BasicGenServer, :start_link, [0]}})
{:ok, child3} = DynamicSupervisor.start_child(sup, %{id: :child3, start: {BasicGenServer, :start_link, [0]}})

:ok = Process.send(child3, :stop_normal, [])
IO.puts("Sent :stop_normal message to child of TestSup")
children = DynamicSupervisor.which_children(sup)

if Enum.count(children, fn {_id, _pid, _type, mod} -> [BasicGenServer] == mod end) != 3 do
  raise "Not 3 BasicGenServers for TestSup"
else
  IO.puts("3 children for DynamicSTestSupupervisor!")
end
```

But running the exact same test with `DynamicSupervisor` replaced with `Horde.DynamicSupervisor` crashes as soon as we send the `:stop_normal` message (i.e. "Sent :stop_normal message to child of TestHordeSup" is never printed):
```
{:ok, sup} = Horde.DynamicSupervisor.start_link([strategy: :one_for_one, name: TestHordeSup, automatic_shutdown: :never])

{:ok, child1} = Horde.DynamicSupervisor.start_child(sup, %{id: :child1, start: {BasicGenServer, :start_link, [0]}})
{:ok, child2} = Horde.DynamicSupervisor.start_child(sup, %{id: :child2, start: {BasicGenServer, :start_link, [0]}})
[...]

Issue #267, opened 2024-02-14
Registry.lookup doesn't always return new item that was just registered


jweinkam:
I have in my application start 
```elixir
{Horde.Registry, members: :auto, keys: :unique, name: MyRegistry}
```
For the code
```elixir
Horde.Registry.register(MyRegistry, "name", "value")
result = Horde.Registry.lookup(MyRegistry, "name") 
```
I would expect result to be [{#PID<...>, "value"}]
But most of the time I get []

jweinkam:
This is using v0.9.0.  Version v0.8.7 always gave the expected result.  I believe this is related to the change in https://github.com/derekkraan/horde/commit/053e1aea33dc19421b175071b59bc615eec7225b

derekkraan:
This is intended behaviour. In local cases, it is expected that you will know the pid of the process (that you have just registered). Is this possible in your case? (Otherwise I would like to know more details)

jweinkam:
I am attempting to do 
```elixir
    GenStateMachine.start_link(__MODULE__, args,  name: {:via, Horde.Registry, {MyRegistry, "name"}})
```
This is frequently failing with error message
```
{:error, {:process_not_registered_via, Horde.Registry}}
```
Since internally in GenStateMachine.start_link it does a register, followed immediately by lookup and get [] instead of [{#PID<...>, ...}]

derekkraan:
Thanks for the clarification. Usually I would expect that you wouldn't register and then immediately do a lookup, since theoretically to register you need the PID anyways, but it seems that `:gen_statem`(?) is doing exactly that.

Do you have a link to the location in the source where this is happening?

I don't have an immediate solution to your problem. I would suggest for now to stick with v0.8.7.

arjan:
Maybe we should let register block until the pid has propagated in the crdt? Although that might make things slow if the sync interval is high.. 🤔

derekkraan:
Calling `Horde.Registry.register/3` does a GenServer call into `Horde.RegistryImpl`, which calls into `DeltaCrdt.put/4`, which results in a `GenServer.call/3` to the DeltaCrdt instance. Before it returns, DeltaCrdt does a few `Kernel.send/2`, back to the `Horde.RegistryImpl`.

So in theory, these messages have all been sent before `Horde.Registry.register/3` returns, which _should_ mean that they are received by `Horde.RegistryImpl` _before_ the next call.

It is confusing to me that this appears not to be the case. I thought that sent messages were always received in order.

[...]

PR #271, opened 2024-06-23
added explanatory note to Horde.DynamicSupervisor

gmdorf:
@derekkraan This is a PR updating the docs to clarify the situation wrt using names vs pids (https://github.com/derekkraan/horde/issues/269#issuecomment-2178012113)

### DIFF
--- a/lib/horde/dynamic_supervisor.ex
+++ b/lib/horde/dynamic_supervisor.ex
@@ -49,6 +49,20 @@ defmodule Horde.DynamicSupervisor do
   ```
 
   Then you can use `MySupervisor.child_spec/1` and `MySupervisor.start_link/1` in the same way as you'd use `Horde.DynamicSupervisor.child_spec/1` and `Horde.DynamicSupervisor.start_link/1`.
+
+  ## Warning: Names and pids are not interchangeable
+
+  In general, the API for Horde.DynamicSupervisor works the same way as the stdlib DynamicSupervisor.
+
+  Howerver, one major difference is that the pid returned by the `start_link` function is not interchangeable with the name of the supervisor.
+
+  This is because a Horde DynamicSupervisor is actually composed of a set of processes that perform necessary functionality, whereas a standard DynamicSupervisor is only one process.
+
+  So, always use the name of the Horde.DynamicSupervisor as the input to functions in this module, and not the pid.
+
+  Using the pid directly will cause undefined behaviour.
+
+  See here for more info: https://github.com/derekkraan/horde/issues/269
   """
   use Supervisor
 


derekkraan:
Thanks for this. I want to improve a bit on this before it goes into main, so I'll hold off on merging it until I have time to make the required changes.

gmdorf:
Yeah, I was thinking that it would also be good to clarify the status of `:via` tuples. AFAIK they don't work?

My boss and I are also scratching our heads trying to figure out if it's possible to nest Horde.DynamicSupervisor. No luck so far.

derekkraan:
It doesn't make sense from my perspective to nest Horde.DynamicSupervisor. My advice would be that if you are considering that, you should likely head back to the drawing board.

I'm not sure what the status is on `:via` tuples for `Horde.DynamicSupervisor`, but they should work for children of `Horde.DynamicSupervisor`. There is the caveat that lookups are eventually consistent, until we figure out a way to fix that for local lookups. (See the people struggling with this in #267).

If you're asking about `:via` tuples for naming `Horde.DynamicSupervisor`, it sounds like this might be related to nesting, and so my advice would be the same as above: consider if this is really the best solution to your problem.

derekkraan:
Horde.DynamicSupervisor is spreading your processes (or sub-trees of processes) evenly across your nodes. What could be the benefit of doing this twice?

Re: adding and removing dynamically at runtime, I would likely just run all of them all the time. An idling Horde.DynamicSupervisor doesn't take up much resources.

gmdorf:
> Horde.DynamicSupervisor is spreading your processes (or sub-trees of processes) evenly across your nodes. What could be the benefit of doing this twice?
Ah, I see.

> 
Re: adding and removing dynamically at runtime, I would likely just run all of them all the time. An idling Horde.DynamicSupervisor doesn't take up much resources.

It's not about conserving resources. It's about not knowing which Horde.DynamicSupervisors will be needed. We have an interface by which the user of our custom IDE can add new Horde.DynamicSupervisors as desired.

derekkraan:
In this case I would see if it would also solve your problem to run it all under a single `Horde.DynamicSupervisor`.

The issue with starting and stopping them at runtime is that in order to have an actual benefit from `Horde.DynamicSupervisor`, you need it to be running on multiple nodes. If you do this dynamically, then you add a bunch of issues. What happens when a new node joins the cluster for example? Or leaves and then comes back, how do you ensure that the list of `Horde.DynamicSupervisor`s are coordinated correctly?

Nesting `Horde.DynamicSupervisor` doesn't help you here, since each child `Horde.DynamicSupervisor` would be run on a single node. As I mentioned above, it needs to run on multiple nodes, otherwise you would be better just nesting a regular `DynamicSupervisor` as a child to your `Horde.DynamicSupervisor`.

gmdorf:
> just nesting a regular DynamicSupervisor as a child to your Horde.DynamicSupervisor.
This seems to make the most sense for my use case.

Thanks Derek! You've been incredibly helpful and made a groundbreaking piece of technology free to use besides!

sleipnir:
Thanks a lot for the work you invested in this! We haven’t been able to give this PR the proper attention, and it has been inactive for a while, so we’re closing it to keep the queue manageable. If you’d like to continue the discussion later, you’re welcome to reopen it anytime. We value your contribution!

PR #274, opened 2024-08-21
feat(DynamicSupervisor): respect `extra_arguments` configuration

noaccOS:
Pass the `extra_arguments` flag to the `ProcessSupervisor`.

Similar to #238


Annopaolo:
@derekkraan any chance this gets merged?

derekkraan:
Yes, thanks for getting my attention with this. I have merged and removed the flags in a follow-on commit from where they were not supposed to be.

PR #278, opened 2025-05-05
v0.8.3 + v0.8.6 fix

maciej-pawlak:


sleipnir:
@maciej-pawlak  is this no longer necessary?

maciej-pawlak:
That's right. I made that by mistake

sleipnir:
> That's right. I made that by mistake

No problem! Thanks for the quick reply

PR #279, opened 2025-06-10
Race condition during Registry.lookup/2

Morgahl:
This addresses an unlikely, but annoyingly consistent during some recent local testing, race condition in `Registry.lookup/2` wherein the target node of the pid disconnects between the call to `Node.list/0` and `:erpc.call/4` resulting in an exception of `:error, {:erpc, :noconnection}`.

I've chosen to only handle the specific case mentioned and otherwise let all other exception behaviour to continue as this specific case at least appears it should return `false`.


derekkraan:
Seems reasonable to me, thanks!

PR #287, opened 2026-03-13
Add typespecs to internal modules and quorum tests

dylantirandaz:
## Summary

- Added `@type t` to `RegistryImpl.State` struct for better static analysis
- Added `@spec` and `@impl` annotations to `SignalShutdown` callbacks
- Added unit tests for `UniformQuorumDistribution` edge cases: empty members, all shutting_down, majority alive vs dead, and quorum_not_met error path

## Test plan

- [x] `mix compile` passes with no warnings
- [x] All quorum distribution tests pass (1 property, 5 tests)

### DIFF
--- a/lib/horde/registry_impl.ex
+++ b/lib/horde/registry_impl.ex
@@ -7,6 +7,18 @@ defmodule Horde.RegistryImpl do
 
   defmodule State do
     @moduledoc false
+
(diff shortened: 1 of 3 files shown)


sleipnir (approved):

sleipnir:
Thank you @dylantirandaz

PR #289, opened 2026-04-06
Add Pluggable Transports and support for Partisan

sahilpohare:
## What this solves

  Horde assumes **Erlang distribution** throughout — `Node.list()` for peer discovery, `:erpc` for remote
  calls, and a single hardcoded `Horde.NodeListener` for membership. This makes Horde incompatible with
  alternative transports like **Partisan**, which replaces Erlang distribution entirely with its own mesh
  networking layer.

  ## What it introduces

  ### `Horde.NodeListenerBehaviour`

  A `use`-based behaviour for building custom node listeners. The `use` macro injects a full GenServer
  implementation with sensible defaults — Erlang distribution monitoring, `handle_nodeup`/`handle_nodedown`,
   and `set_members`. The only required callback is `make_members/1`. All injected functions are
  `defoverridable`.

  `Horde.Registry` and `Horde.DynamicSupervisor` now accept `{:auto, ListenerModule}` as the `:members`
  option to start a custom listener instead of the default `Horde.NodeListener`.

  ### `Horde.ClusterTransport` behaviour

  Abstracts the primitives Horde needs from the underlying cluster transport:

  | Callback | Purpose |
  |---|---|
  | `members/0` | Returns connected peer nodes |
  | `process_alive?/1` | Checks whether a remote pid is alive |
  | `call/5` | Invokes a function on a remote node |

  Two implementations ship out of the box:

  | Module | Backend |
  |---|---|
  | `Horde.ClusterTransport.Erlang` | Standard Erlang distribution (default) |
  | `Horde.ClusterTransport.Partisan` | Partisan — `:partisan_peer_service` + `:partisan_rpc` |

  The `:transport` option on `Horde.Registry` selects the implementation. The chosen module is stored in
  registry ETS metadata at startup and read on every `lookup/2` — one local ETS read, no GenServer call on
  the hot path.

  ### `Horde.NodeListener.Partisan`

  A ready-made Partisan node listener built on `Horde.NodeListenerBehaviour`. Uses
  `:partisan.monitor_nodes/1` for monitoring and `:partisan_peer_service` for membership.

  ```elixir
  {Horde.Registry,
    name: MyRegistry,
    keys: :unique,
    members: {:auto, Horde.NodeListener.Partisan},
    transport: Horde.ClusterTransport.Partisan}
  ```

  ### Conditional compilation

  `Horde.ClusterTransport.Partisan` and `Horde.NodeListener.Partisan` are wrapped in
  `Code.ensure_loaded?(:partisan_peer_service)` guards — only compiled when Partisan is present. No optional
   dep entry in `mix.exs`, no compile-time warnings in Partisan-free builds.

  ## Methodology

  - **No breaking changes** — all defaults preserve existing behaviour. `:transport` defaults to
  `Horde.ClusterTransport.Erlang`, `{:auto, ListenerModule}` is additive alongside the existing `:auto`.
  - **Hot path stays fast** — transport is resolved once at init and stored in ETS metadata, never changes
  at runtime.
  - **Transport is stateless** — just a module, no process or config storage beyond the single ETS write at
  startup.
  - **Existing tests pass** — 104 tests pass. 3 pre-existing failures in
  `ClusterTest`/`NetworkPartitionTest` are unrelated flaky distributed tests.

### DIFF
--- a/lib/horde/node_listener/partisan.ex
+++ b/lib/horde/node_listener/partisan.ex
@@ -1,0 +1,54 @@
+defmodule Horde.NodeListener.Partisan do
+  @moduledoc """
+  A Horde node listener for Partisan-based clusters.
+
+  Uses `:partisan.monitor_nodes/1` for node monitoring and
+  `:partisan_peer_service` for membership, replacing the default
+  Erlang distribution assumptions in `Horde.NodeListenerBehaviour`.
+
+  ## Usage
+
+      {Horde.Registry,
+        name: MyRegistry,
+        keys: :unique,
+        members: {:auto, Horde.NodeListener.Partisan}}
+  """
+
+  use Horde.NodeListenerBehaviour
+
+  @impl GenServer
+  def init(cluster) do
+    :partisan.monitor_nodes(true)
+    {:ok, cluster}
+  end
+
+  @impl Horde.NodeListenerBehaviour
+  def make_members(cluster) do
+    case :partisan_peer_service.members() do
+      members when is_list(members) ->
+        Enum.map(members, fn peer -> {cluster, peer.name} end)
+
+      _ ->
+        [{cluster, :partisan_config.get(:name)}]
+    end
+  end
+
+  @impl Horde.NodeListenerBehaviour
+  def handle_nodeup(_node, cluster), do: set_members(cluster)
+
+  @impl Horde.NodeListenerBehaviour
+  def handle_nodedown(_node, cluster), do: set_members(cluster)
+
+  # Partisan emits 2-tuple {:nodeup, node} / {:nodedown, node} — no node_type
+  @impl GenServer
+  def handle_info({:nodeup, node}, cluster) do
+    handle_nodeup(node, cluster)
+    {:noreply, cluster}
+  end
+
+  @impl GenServer
+  def handle_info({:nodedown, node}, cluster) do
+    handle_nodedown(node, cluster)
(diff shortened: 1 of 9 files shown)


mix.exs
```diff
@@ -14,7 +14,7 @@ defmodule Horde.MixProject do
       docs: docs(),
       package: package(),
       name: "Horde",
-      source_url: "https://github.com/elixir-horde/horde",
+      source_url: "https://github.com/derekkraan/horde",
```
sleipnir: The correct is https://github.com/elixir-horde/horde

sleipnir (changes requested):

mix.exs
sahilpohare: Updated

sleipnir:
Hi @sahilpohare, thanks for the PR, I'll review it soon.

lib/horde/node_listener/partisan.ex
```diff
@@ -0,0 +1,54 @@
+defmodule Horde.NodeListener.Partisan do
```
sleipnir: Would it be a good idea to wrap this under a `Code.ensure_loaded`?

sleipnir (changes requested):

lib/horde/node_listener/partisan.ex
sahilpohare: Done

sleipnir (approved):

